PR #5, opened 2021-01-20
Bump version to 1.0 and fix up Debian packaging a bit

geofft:
- Add missing pkg-config dep so libsystemd_sys is happy
- Use [patch] mechanism to work around dpkg-source not putting
  .gitignore files in the source tarball
- Expand docs a bit
- Avoid Lintian errors (fix d/copyright and ignore missing sources i
  vendor/)
- Bump version number to 1.0 (to supersede previous internal builds)

### DIFF
--- a/debian/README.source
+++ b/debian/README.source
@@ -1,12 +1,24 @@
-Since Debian Stretch does not have a complete Rust packaging environment
-yet, we build the Debian package by first using "cargo vendor" to
-downlooad the dependencies. To run this automatically and set things up,
-run
+Because Debian Stretch does not have a complete Rust packaging
+environment yet, we build the Debian package by first using "cargo
+vendor" to download the dependencies. To run this automatically and set
+things up, manually run
 
     debian/rules vendor
 
-You can then build with dpkg-buildpackage (or commit and build with gbp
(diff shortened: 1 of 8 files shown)


debian/README.source
```diff
@@ -1,12 +1,24 @@
-Since Debian Stretch does not have a complete Rust packaging environment
-yet, we build the Debian package by first using "cargo vendor" to
-downlooad the dependencies. To run this automatically and set things up,
-run
+Because Debian Stretch does not have a complete Rust packaging
+environment yet, we build the Debian package by first using "cargo
+vendor" to downlooad the dependencies. To run this automatically and set
```
leifwalsh, line 3: s/downlooad/download/

debian/changelog
```diff
@@ -1,5 +1,5 @@
-nsncd (0.1) unstable; urgency=medium
+nsncd (1.0) unstable; urgency=medium
```
leifwalsh, line 1: let's bump the version number in Cargo.toml too?

leifwalsh (changes requested):

leifwalsh (approved):

PR #8, opened 2021-03-02
handle keys with internal nuls and ignore systemd

### LINKED ISSUE #7: Confusing log messages about nul termination
If you run e.g. `getent hosts 127.0.0.1`, that causes a GETHOSTBYNAME request looks something like

    02 00 00 00  06 00 00 00  04 00 00 00  7f 00 00 01

which causes nsncd to log "data provided contains an interior nul byte at byte pos 1" and drop the connection.

If you run e.g. `getent hosts 127.127.127.127`, that causes the request

    02 00 00 00  06 00 00 00  04 00 00 00  7f 7f 7f 7f

which logs "data provided is not nul terminated".

I think this is coming from the `CStr::from_bytes_with_nul` call in `Request::parse`, and I think we should not try to parse the argument as a NUL-terminated bytestring until we know that it's actually an operation that uses a NUL-terminated bytestring.

(This isn't a behavior bug, because we only support operations that use NUL-terminated bytestrings anyway, but it'd be nice to make the log spam go away - in particular every incoming ssh connection causes a reverse DNS lookup :) )

leifwalsh:
Deserialize keys that have nuls in them (because these might be ip
addresses), handle nul-terminated strings higher up when we know what
kind of request it is.

Don't use systemd socket activation becuase startup ordering is weird.

fixes #7

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -15,7 +15,7 @@
  */
 
 use std::convert::TryInto;
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::os::unix::ffi::OsStrExt;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -83,13 +83,11 @@ fn main() -> Result<()> {
...
-        listen_fds_found == 1,
-        "expected one listen fd, got {}",
-        listen_fds_found
-    );
-    let listener = unsafe { UnixListener::from_raw_fd(LISTEN_FDS_START) };
+    let path = Path::new(SOCKET_PATH);
+    std::fs::create_dir_all(path.parent().expect("socket path has no parent"))?;
+    std::fs::remove_file(path).ok();
+    let listener = UnixListener::bind(path).context("could not bind to socket")?;
+    std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o777))?;
```
geofft, line 90:
```suggestion
    std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o666))?;
```

src/main.rs
```diff
@@ -83,13 +83,11 @@ fn main() -> Result<()> {
...
-    let listen_fds_found = listen_fds(true)?;
-    ensure!(
-        listen_fds_found == 1,
-        "expected one listen fd, got {}",
-        listen_fds_found
-    );
-    let listener = unsafe { UnixListener::from_raw_fd(LISTEN_FDS_START) };
+    let path = Path::new(SOCKET_PATH);
+    std::fs::create_dir_all(path.parent().expect("socket path has no parent"))?;
+    std::fs::remove_file(path).ok();
```
geofft, line 88: isn't `let _ =` more idiomatic? (I am not actually sure)

geofft (commented):
LGTM but the systemd stuff also needs changes to the unit files, want to just split them into separate PRs?

src/main.rs
leifwalsh: `let _ =` what?

leifwalsh:
Yeah I intentionally didn't touch the unit files, I sort of wanted you to do that. If you can guide my hand I can do it here.

src/main.rs
leifwalsh: I think `std::fs::remove_file(path).ok()` is equivalent to saying "remove the file if possible, and I expect that if there are any errors with that we should proceed, including the one I expect most which is `ENOENT` which is fine, any other errors like permissions will probably surface when we try to bind".

geofft (approved):

src/main.rs
geofft: I'm used to seeing `let _ = std::fs::remove_file(path)`. But apparently silencing the warning with `.ok()` is also reasonable according to https://stackoverflow.com/questions/51141672/how-do-i-ignore-an-error-returned-from-a-rust-function-and-proceed-regardless

PR #11, opened 2021-03-10
add simple devcontainer setup

leifwalsh:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.154.0/containers/rust
+{
+	"name": "Rust",
+	"build": {
+		"dockerfile": "Dockerfile"
```
blinsay, line 6: Can this be `image: mcr.microsoft.com/vscode/devcontainers/rust:0-1`? Seems like that would skip the first build.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+		"lldb.executable": "/usr/bin/lldb",
+		// VS Code don't watch files under ./target
+		"files.watcherExclude": {
+			"**/target/**": true
+		}
+	},
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"rust-lang.rust",
```
blinsay, line 22: Rust analyzer? 🙏 

https://marketplace.visualstudio.com/items?itemName=matklad.rust-analyzer

blinsay (commented):
This is nice!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.154.0/containers/rust
+{
+	"name": "Rust",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
```
geofft, line 8: You shouldn't need CAP_SYS_PTRACE with seccomp=unconfined, but maybe it allows you to use an unprivileged lldb to debug an nsncd running as root?

leifwalsh: This is the default rust config vscode gave me. I didn't add these, but I can remove them if you think we should.

geofft: lol, okay, leave it and I'll file a bug report against vscode :)

.devcontainer/devcontainer.json
leifwalsh: idk I took the defaults

.devcontainer/devcontainer.json
leifwalsh: https://github.com/microsoft/vscode-remote-try-rust/blob/master/.devcontainer/devcontainer.json#L8

PR #12, opened 2021-03-13
Add some basic tests

blinsay:
Adds some basic tests.

`protocol.rs` tests are there to verify that `repr(C)` is doing what we expect and producing some sane layout. If it's not, we might be reading garbage from libc nscd. The manual version ended up being easier to read and think about a week later than the version using const asserts.

`handlers.rs` are a starting point for more handlers test. I'd like to test our error handling, but `anyhow` errors don't implement `Eq` so it's a little bit annoying to assert they're doing the right thing.

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -172,3 +172,70 @@ where
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    fn send_to(v: &mut Vec<u8>) -> impl FnMut(&[u8]) -> Result<()> + '_ {
+        move |bs| {
+            v.extend_from_slice(bs);
+            Ok(())
+        }
+    }
+
+    fn test_logger() -> slog::Logger {
+        Logger::root(slog::Discard, slog::o!())
+    }
+
+    #[test]
+    fn test_handle_request_empty_key() {
+        let mut output = vec![];
+        let request = protocol::Request {
+            ty: protocol::RequestType::GETPWBYNAME,
+            key: &[],
+        };
+
+        assert!(
+            handle_request(&test_logger(), &request, send_to(&mut output)).is_err(),
+            "should error on empty input"
+        );
+        assert!(output.is_empty());
+    }
+
+    #[test]
+    fn test_handle_request_nul_data() {
+        let mut output = vec![];
+        let request = protocol::Request {
+            ty: protocol::RequestType::GETPWBYNAME,
+            key: &[0x7F, 0x0, 0x0, 0x01],
+        };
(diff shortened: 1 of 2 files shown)


src/handlers.rs
```diff
@@ -172,3 +172,72 @@ where
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    macro_rules! send_to {
```
leifwalsh, line 180: Just for my own curiosity, why is this a macro instead of a function that returns a closure?

src/handlers.rs
```diff
@@ -172,3 +172,72 @@ where
...
+        };
+
+        let mut expected = vec![];
+        send_user(&test_logger(), Some(current_user), send_to!(expected))
+            .expect("send_user should serialize current user data");
+
+        let mut output: Vec<u8> = vec![];
+        handle_request(&test_logger(), &request, send_to!(output))
+            .expect("should handle request with no error");
+        assert_eq!(expected, output,);
```
leifwalsh, line 241: Extra comma?

src/handlers.rs
```diff
@@ -172,3 +172,72 @@ where
...
+        );
+        assert!(output.is_empty());
+    }
+
+    #[test]
+    fn test_handle_request_nul_data() {
+        let mut output = vec![];
+        let request = protocol::Request {
+            ty: protocol::RequestType::GETPWBYNAME,
+            key: &[0x7F, 0x0, 0x0, 0x01],
```
leifwalsh, line 213: Is there something special about these bytes?

leifwalsh (approved):

src/handlers.rs
blinsay: weird, that survived `rustfmt`. I'll remove it.

src/handlers.rs
blinsay: I picked something with an interior null byte just for fun (`127.0.0.1`) but anything that's not a username should work here.

src/handlers.rs
blinsay: I found the macro easier to do the first time, but the fn makes sense too. Happy to do whatever here, the diff is real small.

```rust
    fn send_to(v: &mut Vec<u8>) -> impl FnMut(&[u8]) -> Result<()> + '_ {
        move |bs| {
            v.extend_from_slice(bs);
            Ok(())    
        }
    }
```

leifwalsh: Meh, either way.

leifwalsh (approved):

PR #16, opened 2021-03-23
add context information to numeric conversions

leifwalsh:


blinsay:
This feels like a good point to stop and figure out how to do some tests on Anyhow errors or wire up a custom error with slightly more data than a `&str`. Time for thiserror?

PR #20, opened 2021-04-08
add clippy and ci test to GHA

leifwalsh:


geofft:
I don't remember why we did debhelper 12 exactly but chances are high it was for systemd support. (See the difference between compat levels in https://manpages.debian.org/testing/debhelper/debhelper.7.en.html .) We could downgrade it to 10 if it makes life easier for CI, it'd probably make life easier for other people too.

geofft:
Huh. Clippy is still returning warnings, but GHA isn't noting that as fatal.

PR #21, opened 2021-04-08
Make handle_request take a &mut Write instead of a closure

geofft:
In normal use, this is a UnixStream; in tests, this is a Vec<u8>.

Seems a bit easier to follow the logic.


geofft:
Not sure if we want to do this or just have the handler return a Vec<u8> (or even a Vec<Vec<u8>>, honestly, which I think would be fine and would continue to avoid the extra copy).

geofft:
Oh I missed #19, which conflicts with this.

blinsay:
wanna close in favor of #19?

PR #23, opened 2021-04-11
add direct tests of serialization functions

leifwalsh:


src/handlers.rs
```diff
@@ -213,4 +214,119 @@ mod test {
...
+        expected.extend_from_slice(&c_int::from(0i32).to_ne_bytes());
+        // pub pw_shell_len: c_int,
+        expected.extend_from_slice(&c_int::from(0i32).to_ne_bytes());
+        let output = serialize_user(None).unwrap();
+        assert_eq!(expected, output);
+    }
+
+    #[test]
+    fn test_serialize_user() {
+        let user = User::from_name("root").unwrap().unwrap();
```
blinsay, line 245: can we construct a `User` here instead of asking libc for one?

```rust
let fakey_mc_fakerson = User{name: "fake".to_string(), ... };
```

blinsay (commented):
Torn between this and having a snapshot-test-style thing where we actually write out a literal response instead of constructing it again in the test. We'd have to `[cfg(target_endian)]` a be and a le value too - painful.

blinsay:
> Torn between this and having a snapshot-test-style thing where we actually write out a literal response instead of constructing it again in the test. We'd have to `[cfg(target_endian)]` a be and a le value too - painful.

Took a stab at it just to see how it felt. Still not sure.

https://github.com/twosigma/nsncd/compare/benl/test-serialization-alt

leifwalsh:
Pretty sure we decided against this and just forgot to close it out, if my memory is fuzzy we can pick it back up later.

Issue #24, opened 2021-05-15
"fork rejected by pids controller" message


geofft:
On an internal machine, I happened to notice the following message in dmesg repeated a few times:

```
cgroup: fork rejected by pids controller in /system.slice/nsncd.service
```

and nsncd had died with SIGABRT.

I don't have a sense of what caused this issue (this is a shared dev box), and I wasn't able to trivially reproduce it. On my workstation, /sys/fs/cgroup/pids/system.slice/nsncd.service/pids.max is 4915, and I tried making 5000 `getent passwd geofft &` requests in parallel and I wasn't able to trigger it. But I didn't try very hard.

There's also probably more logging on the box itself, which I haven't gone looking for. It may be notable that the box also has a pile of OOM kills from user code.

Still, we should probably, at the last, stop accepting new connections if we have more than, say, ~1000 children, and maybe hold onto the connection for a brief timeout and drop it.

PR #25, opened 2021-05-20
Limit the number of handler threads

blinsay:
Adds a semaphore and uses it to limit the total number of threads spawned when handling requests.

This is a potential fix for #24. @geofft suggested testing both the bug and this fix by writing an nss module that does `sleep 10` on every request and issuing a ton. I haven't gotten around to that yet, but still wanted to put this up.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,6 +50,7 @@ use std::os::unix::fs::PermissionsExt;
 use std::os::unix::net::{UnixListener, UnixStream};
 use std::path::Path;
 use std::thread;
+use std::time::Duration;
 
 use anyhow::{Context, Result};
 use slog::{debug, error, o, Drain};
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -50,13 +50,65 @@ use std::os::unix::fs::PermissionsExt;
...
 mod ffi;
 mod handlers;
 mod protocol;
+mod semaphore;
+
+const SOCKET_PATH: &str = "/var/run/nscd/socket";
+const MAX_THREADS: usize = 64;
+const MAX_THREADS_WAIT_TIMEOUT: Duration = Duration::from_secs(5);
+
+fn main() -> Result<()> {
```
blinsay, line 67: I just noticed I reordered `main` while moving `Semaphore` to its own module. I can leave it or revert that.

src/main.rs
```diff
@@ -50,13 +50,65 @@ use std::os::unix::fs::PermissionsExt;
...
+        match stream {
+            Ok(stream) => {
+                let thread_logger = logger.clone();
+
+                // block on acquring a permit before spawning the thread so
+                // that we actually limit the number of threads used.
+                let permit = match sem.acquire(MAX_THREADS_WAIT_TIMEOUT) {
+                    Ok(p) => p,
+                    Err(_) => {
+                        anyhow::bail!("timed out waiting to spawn a handler thread. blowing up!")
```
blinsay, line 94: @leifwalsh @geofft this should be a better error message. Thoughts on what it should be?

src/semaphore.rs
```diff
@@ -0,0 +1,125 @@
...
+// clippy does a false alarm if you're trying to use a Condvar.
+//
+// https://rust-lang.github.io/rust-clippy/master/index.html#mutex_atomic
+// https://github.com/rust-lang/rust-clippy/issues/1516
+#![allow(clippy::mutex_atomic)]
+
+use std::sync::{Arc, Condvar, Mutex};
+use std::time::Duration;
+
+/// A counting semaphore.
```
leifwalsh, line 26: Is there no semaphore crate we can just use? I'm a little surprised we need to roll our own here.

src/main.rs
leifwalsh: Is there a reason you went with limiting the number of threads with a semaphore, but still forking on each request, instead of preforking a threadpool of that many threads and bailing if we can't get one of them to accept our request within the timeout?

blinsay: I remember having a conversation about it in slack, but can't remember why we decided on this. I can put up a PR with a [threadpool](https://crates.io/crates/threadpool) tonight so we can compare.

src/semaphore.rs
blinsay: The ones that [show up on crates.io](https://crates.io/search?q=semaphore) look like they're really low usage. This felt better than taking a dependency on something unowned/unmaintained.

blinsay:
Closed in favor of #28

PR #26, opened 2021-05-24
Add initgroups support

geofft:
Apart from its own inherent value in rounding out user/group support,
glibc uses the same variable to cache whether the current NSCD
implementation handles group (GETGRBYNAME/GETGRBYGID) and INITGROUPS
requests, so we need to have a working implementation of it.

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -20,8 +20,8 @@ use std::os::unix::ffi::OsStrExt;
 
 use anyhow::{Context, Result};
 use atoi::atoi;
-use nix::unistd::{Gid, Group, Uid, User};
-use slog::{debug, Logger};
+use nix::unistd::{getgrouplist, Gid, Group, Uid, User};
+use slog::{debug, error, Logger};
 
 use super::protocol;
 use super::protocol::RequestType;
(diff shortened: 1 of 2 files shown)


src/handlers.rs
```diff
@@ -62,6 +62,39 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
...
+            // if you ask for a nonexistent user, they just happen not to be in
+            // any groups. So the "found" value is mostly used to indicate
+            // whether the running NSCD knows about and is configured to serve
+            // initgroups responses. For that reason, we return found = 1 even
+            // if User::from_name fails.
+            let key = CStr::from_bytes_with_nul(request.key)?;
+            let user = User::from_name(key.to_str()?)?;
+            debug!(log, "got user"; "user" => ?user);
+            let groups = if let Some(user) = user {
+                getgrouplist(key, user.gid)?
```
blinsay, line 92: the comment above says to always return success, but this line implies it bails early if something goes wrong. looks like `nix` adds a too-many-groups error condition, even though libc is never going to return an error.

https://docs.rs/nix/0.20.0/nix/unistd/fn.getgrouplist.html

blinsay: might be less confusing to pull this if/else into `serialize_initgroups` and change than fn so it takes the username and primary gid

geofft: 1. I changed this specific line to log that error condition and return an empty list, and I expanded the comment to explain why we return an empty list.
2. There are still error conditions in `serialize_initgroups` that we don't handle. It's potentially important that we don't drop the connection there. However, that's a problem for _all_ of the `serialize_` functions, i.e, I think we should consider teaching `main()` not to drop the connection but instead to provide some perfunctory reply and log what's going on, because all of those cases can cause glibc to decide that there isn't a working NSCD. I think that's for another PR, and we shouldn't special-case `serialize_initgroups` here. (I also think that several of these cases can be statically proven not to be problems, if we try harder, since they're integer conversions.)

blinsay: > However, that's a problem for all of the serialize_ functions, i.e, I think we should consider teaching main() not to drop the connection but instead to provide some perfunctory reply and log what's going on, because all of those cases can cause glibc to decide that there isn't a working NSCD. I think that's for another PR, and we shouldn't special-case serialize_initgroups here.

+1

blinsay (approved):
this all lgtm

PR #27, opened 2021-06-29
Use a threadpool to handle incoming requests

### LINKED ISSUE #24: "fork rejected by pids controller" message

blinsay:
Adds a pool of worker threads to handle incoming requests. Uses a bounded channel to pass incoming requests to each worker. Closes #24.

### Alternatives

This is an alternative to #25 which uses a semaphore to limit the number of requests in flight but still spawns a thread per request. This version uses existing dependencies instead of rolling our own semaphore.

There's another (easy) alternative to this approach that doesn't use a channel to pass the incoming UnixStream to worker threads:

```rust
    let pool = ThreadPool::new(N_WORKERS);

    for stream in listener.incoming() {
        match stream {
            Ok(stream) => {
                let logger = logger.clone();
                pool.execute(move || handle_stream(logger, stream));
            }
            Err(err) => {
                error!(logger, "error accepting connection"; "err" => %err);
                break;
            }
        }
    }
```

Calling `pool.execute` like this doesn't apply any backpressure. `ThreadPool` is build on `std::sync::mpsc` unbuffered channels so `pool.execute` [never blocks](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#method.send). Using this means we have no signal we can use to do something (log? crash the process?) if things get slow.

On the plus side, this approach is less code and handles panics gracefully - ThreadPool makes sure there are always N_WORKERS threads in the pool. With the approach in the PR, on panic we'll kill a worker thread but nothing starts a new worker task.

I don't have a strong preference between the threadpool approaches, but they both seem better than #25. The `pool.execute` version requires the least code but doesn't let us detect a hang and crash.


Cargo.toml
```diff
@@ -18,9 +18,11 @@ atoi = "^0.3"
 slog = "^2.5"
 slog-async = "^2.5"
 slog-term = "^2.6"
+crossbeam-channel = "^0.4"
```
blinsay, line 21: This is the same version of crossbeam-channel we pull in through slog-async.

src/main.rs
```diff
@@ -49,15 +49,95 @@ use std::io::ErrorKind;
...
+                // giving up and going elsewhere so crashing the process should
+                // not make a bad situation worse.
+                match handle.send_timeout(stream, HANDOFF_TIMEOUT) {
+                    Err(channel::SendTimeoutError::Timeout(_)) => {
+                        anyhow::bail!("timed out waiting for an available worker: exiting")
+                    }
+                    Err(channel::SendTimeoutError::Disconnected(_)) => {
+                        anyhow::bail!("aborting: worker channel is disconnected")
+                    }
+                    _ => { /*ok!*/ }
```
leifwalsh, line 101: Can we make the last clause just `Ok`? I'm worried about send_timeout starting to return new error types and us silently ignoring them. I'd rather have an exhaustive match block.

src/main.rs
```diff
@@ -49,15 +49,95 @@ use std::io::ErrorKind;
...
+    let (tx, rx) = channel::bounded(0);
+
+    // TODO: figure out how to name the worker threads in each worker
+    // TODO: report actively working threads
+    for _ in 0..n_workers {
+        let log = log.clone();
+        let rx = rx.clone();
+
+        pool.execute(move || loop {
+            let log = log.clone();
```
leifwalsh, line 130: Why do we need another clone here?

src/main.rs
```diff
@@ -49,15 +49,95 @@ use std::io::ErrorKind;
...
+            match rx.recv() {
+                Ok(stream) => handle_stream(log, stream),
+                Err(channel::RecvError) => break,
```
leifwalsh, line 133: Can we log something useful on error?

src/main.rs
```diff
@@ -49,15 +49,95 @@ use std::io::ErrorKind;
...
+
+    ffi::disable_internal_nscd();
+
+    let decorator = slog_term::TermDecorator::new().build();
+    let drain = slog_term::FullFormat::new(decorator).build().fuse();
+    let drain = slog_async::Async::new(drain).build().fuse();
+
+    let logger = slog::Logger::root(drain, slog::o!());
+
+    let (pool, handle) = worker_pool(logger.clone(), N_WORKERS);
```
leifwalsh, line 76: Can we call `handle` `channel` or `tx` or something instead? "handle" is so generic

leifwalsh (commented):
I definitely prefer something with backpressure to the `pool.submit(move || handle_stream(...))` one where some collection of unaccepted tasks grows unbounded out of our view.

I do agree, with the version in this PR, we need some communication back up from threads which errored or panicked to the main thread, possibly crashing the process.

src/main.rs
blinsay: Nope. It's just a sentinel telling us the channel is closed.

https://docs.rs/crossbeam-channel/0.5.1/crossbeam_channel/struct.RecvError.html

blinsay:
Closed in favor of #28

PR #28, opened 2021-07-01
workgroup implementation of bounded threads

leifwalsh:


src/main.rs
```diff
@@ -49,17 +49,132 @@ use std::io::ErrorKind;
...
+                    error!(log, "error accepting connection"; "err" => %err);
+                    break;
+                }
+            }
+        }
+
+        // at the end of the listener loop, drop tx so that any working threads
+        // still waiting for a connection have a chance to finish.
+        //
+        // this is unecessary but explicit
```
leifwalsh, line 148:
```suggestion
        // this is unnecessary but explicit
```

src/main.rs
```diff
@@ -49,17 +49,132 @@ use std::io::ErrorKind;
...
+    n_workers: usize,
+) -> channel::Sender<UnixStream> {
+    let (tx, rx) = channel::bounded(0);
+
+    for worker_id in 0..n_workers {
+        let rx = rx.clone();
+        let log = log.new(o!("thread" => format!("worker_{}", worker_id)));
+
+        // ctx is ignored - the acceptor thread will close the rx channel if
+        // the wg is shutdown and its time to exit.
```
leifwalsh, line 165:
```suggestion
        // the wg is shutdown and it's time to exit.
```

src/work_group.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn is_shutdown(&self) -> bool {
+        let (m, _) = &*self.data;
+        m.lock().unwrap().is_some()
+    }
+}
+
+impl Drop for Context {
+    fn drop(&mut self) {
+        let (mtx, cond) = &*self.data;
+        *mtx.lock().unwrap() = Some(self.thread);
```
leifwalsh, line 25: Should we worry at all about multiple threads exiting before `run` wakes up and returns? This blindly overwrites the "first" thread that exited, so it's sort of LWW. I don't really think that matters, probably? The contract is just that when `run` terminates, it gives you one of the witnesses to an exit/error, and a vec of the remaining threads, which may have already terminated.

leifwalsh: Maybe a richer API would be to have `run` return `(Vec<Result<(), Box<dyn Any + Send + 'static>>>, Vec<JoinHandle<()>>>)`, but even then, some of the remaining threads in the second argument could exit before the caller resumes control.

I guess, what happens if you `join` a `JoinHandle` twice? Do you still get the same panic value or Ok result the second time?

blinsay: > Should we worry at all about multiple threads exiting before run wakes up and returns? This blindly overwrites the "first" thread that exited, so it's sort of LWW.

This is absolutely a bug and should check `is_none()` before overwriting the value. I'll add a test too.

blinsay: > I guess, what happens if you join a JoinHandle twice? Do you still get the same panic value or Ok result the second time?

`join` consumes the `JoinHandle` so you can't call it twice. We could figure out how to return all of the completed ones already, but I think I'm happy to leave that to the caller. The intended usage is to catch a panic in things that should run forever-ish.

src/main.rs
blinsay: pyto fixed

src/main.rs
blinsay: pyto fixed

src/work_group.rs
blinsay: The bug is fixed (the offending assignment [get_or_insert now](https://github.com/twosigma/nsncd/pull/28/files#diff-e6782ce606882da51efbd9c9e71adf99b28862e3b7bfffc1354d438e4dc549b2R25)), but I'm not quite sure how to test it. I think we'd have to reach in to the `Arc` and watch the internal state somehow? Thoughts?

leifwalsh: I can't think of a way to test it but tbh I'm not that worried. I think even if we regressed it would still kinda be okay.

PR #32, opened 2021-10-07
Release 1.2.1

geofft:
[testing the failing test first]


ci/libnss_whatami.c
```diff
@@ -20,11 +20,12 @@
 #include <nss.h>
 #include <pwd.h>
 #include <string.h>
+#include <stdlib.h>
 
 enum nss_status
 _nss_whatami_getpwnam_r(const char *name, struct passwd *result, char *buffer, size_t buflen, int *errnop)
 {
-	if (strcmp(name, "whatami") == 0) {
+	if (strcmp(name, "whatami") == 0 || strncmp(name, "am_i_", 5) == 0) {
```
blinsay, line 28: since we're accepting more names, should we set `result->pw_name = name`? it probably doesn't matter for this test but maybe we won't regret it later.

leifwalsh (approved):
I have...questions...about the implementation of `_nss_whatami_initgroups_dyn`, but also, if the test works it works and maybe that's okay.

Let me know if you'd like me to comment all over that function, or just merge.

PR #33, opened 2021-10-29
Make worker threads and timeouts configurable.

blinsay:
Allow configuring the number of worker threads and handoff timeout with `NSNCD_WORKER_COUNT` and `NSNCD_HANDOFF_TIMEOUT` to tune cpu usage on busy hosts.

This also changes the default number of workers from 256 to 8. We may want to roll that out separately.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -44,6 +44,7 @@
 // - test errors in underlying calls
 // - daemon/pidfile stuff
 
+use std::env;
 use std::io::prelude::*;
 use std::io::ErrorKind;
 use std::os::unix::fs::PermissionsExt;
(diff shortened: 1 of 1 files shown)


leifwalsh (approved):

src/main.rs
```diff
@@ -75,15 +74,24 @@ fn main() -> Result<()> {
 
     let logger = slog::Logger::root(drain, slog::o!());
 
+    let worker_count = env_usize("NSNCD_WORKER_COUNT", 8);
+    let handoff_timeout = Duration::from_secs(env_usize("NSNCD_HANDOFF_TIMEOUT", 3) as u64);
```
domenkozar, line 78: 3 seconds is way too low, at cachix.org we've been experiencing problems due to this. Can we make it 10?

blinsay: Are you able to tune the timeout with NSNCD_HANDOFF_TIMEOUT?

domenkozar: Yes, however I'd much prefer a fix upstream.

https://github.com/NixOS/nixpkgs/pull/290355

blinsay: Can you open an issue where we can discuss this? Tell us more about what kind of load you're putting on nsncd at Cachix, how often you see the timeouts and on which calls, and what your upstream NSS calls are (e.g. are you reaching out to LDAP or some other networked service).

PR #35, opened 2022-10-07
use sd-notify to signal readyness

flokli:
Provide a way to signal nsncd has successfully started up, and is ready to accept connections.

A more reliable way would be to only signal this once at least one worker has started up, but this is good enough (tm) for now.

This can be used in a systemd.service file with Type=notify.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,6 +54,7 @@ use std::time::Duration;
 
 use anyhow::{Context, Result};
 use crossbeam_channel as channel;
+use sd_notify::NotifyState;
 use slog::{debug, error, o, Drain};
 
 mod ffi;
@@ -83,7 +84,6 @@ fn main() -> Result<()> {
         "worker_count" => worker_count,
         "handoff_timeout" => ?handoff_timeout,
     );
-
     let mut wg = WorkGroup::new();
     let tx = spawn_workers(&mut wg, &logger, worker_count);
 
@@ -92,6 +92,8 @@ fn main() -> Result<()> {
     let listener = UnixListener::bind(path).context("could not bind to socket")?;
     std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o777))?;
     spawn_acceptor(&mut wg, &logger, listener, tx, handoff_timeout);
+
+    let _ = sd_notify::notify(true, &[NotifyState::Ready]);
 
     let (result, handles) = wg.run();
     if let Err(e) = result {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -446,6 +446,7 @@ dependencies = [
  "num-derive",
  "num-traits",
  "once_cell",
+ "sd-notify",
  "slog",
  "slog-async",
  "slog-term",
@@ -646,6 +647,12 @@ name = "scratch"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9c8132065adcfd6e02db789d9285a0deb2f3fcb04002865ab67d5fb103533898"
+
+[[package]]
+name = "sd-notify"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "621e3680f3e07db4c9c2c3fb07c6223ab2fab2e54bd3c04c3ae037990f428c32"
 
 [[package]]
 name = "serde"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -650,9 +650,9 @@ checksum = "9c8132065adcfd6e02db789d9285a0deb2f3fcb04002865ab67d5fb103533898"
 
 [[package]]
 name = "sd-notify"
-version = "0.4.1"
+version = "0.3.0"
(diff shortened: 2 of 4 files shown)


leifwalsh:
@flokli can you tell us a bit more about why you want to add this? What can you do with this change that you can't without?

I know this isn't the change you're proposing, but just FYI if you're doing systemd things, we tried using socket activation once and hit a deadlock: systemd only activates one service at a time, so if the activation of another service tries to do a NSS lookup, that will make systemd _enqueue_ the activation of nsncd, but won't actually try to start it until the other service completes its own startup (which then won't proceed since it's waiting for nsncd to respond to the lookup). #10

leifwalsh:
Actually, from #9:

> I think it's also best to avoid readiness notification, because I don't totally understand whether the time between starting nsncd and getting the readiness notification counts as a running job and what that means for the job queue. We should just make nsncd a Type=simple service. This might cause race conditions with services that require nsncd (i.e., services with a foreign libc) that try to say Requires=nsncd.service, because they don't know when the NSCD socket has actually been created. If that causes a problem in practice for anyone, we should come back and add readiness notification and test it a bit more.

So...it sounds like @geofft considered that the change you're proposing might be good, but also maybe could cause similar job queue problems? I think we just never tested this out and if you want it it might be good but we should test carefully.

I'll wait to hear more about your use case first though.

flokli:
> So...it sounds like @geofft considered that the change you're proposing might be good, but also maybe could cause similar job queue problems? I think we just never tested this out and if you want it it might be good but we should test carefully.

> I'll wait to hear more about your use case first though.

I plan using nsncd on NixOS, as a drop-in replacement for glibc-nscd (pending https://github.com/twosigma/nsncd/issues/37). In NixOS, as there's no common `/usr/lib`, glibc code paths in client applications don't know how to load any "non-glibc" nss modules listed in `/etc/nsswitch.conf`, so we absolutely need to do the lookups via the nscd socket - otherwise lookups provided by external NSS modules fail.

We put `nscd.service` with a `RequiredBy=nss-lookup.target nss-user-lookup.target` (and even think about [pulling it in in early boot](https://github.com/NixOS/nixpkgs/pull/154928)), so systemd waits until nscd is ready before starting services that have `After=` set to one of these targets.

glibc-nscd does fork when it's ready (`Type=forking`), but nsncd is currently using `Type=simple`, so systemd has no idea when the service is ready, starts it as soon as it started the nsncd process, while that one might not have created the unix socket, to accept connections. So downstream clients fail to connect.

This change lets nsncd signal readyness via https://docs.rs/sd-notify/latest/sd_notify/ once the nscd unix socket has been created, which is more precise.

I don't see how this could cause any deadlocking of service startup. 

The issue @geofft mentioned was probably caused by the socket activation, and systemd only activating one service at a time (?) [^1]

[^1]: even though that smells like a bug that should be investigated, but let's leave it out of scope for here.

flokli:
With https://github.com/twosigma/nsncd/pull/36 merged, I rebased this.

It now only adds the `sd-notify` to Cargo.lock, and all the upgrade noise is gone.

src/main.rs
```diff
@@ -93,6 +93,8 @@ fn main() -> Result<()> {
     std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o777))?;
     spawn_acceptor(&mut wg, &logger, listener, tx, handoff_timeout);
 
+    let _ = sd_notify::notify(true, &[NotifyState::Ready]);
```
geofft, line 96: Was curious about this but a) that's the [example in the Rust crate](https://docs.rs/sd-notify/latest/sd_notify/) and b) [the C docs](https://www.freedesktop.org/software/systemd/man/sd_notify.html) specifically say, "...it is generally recommended to ignore the return value of this call."

geofft (approved):
Yeah, I think I'm convinced that this cannot deadlock: the problem is that when one job _asks systemd to start up the NSCD_, systemd won't do it until the job is done. Here we're not asking systemd to do anything. We're informing systemd that we're ready, but we're already listening and handling requests on our own. systemd is asking _us_ if we've started up so it can enqueue downstream jobs. Even if the time between process start and readiness notification counts as a job to systemd (which I'm still unclear about) and prevents other jobs from running, there aren't other jobs, i.e., nsncd itself doesn't need to activate anything else.

(And yes, I believe this is a systemd bug or at least design flaw, and I was surprised by the response to that freedesktop bug report linked in #9. But if systemd is designed around a concept of jobs that are serially executed, it seems like a pain to fix, and it seems like the maintainers are saying "You're not supposed to use systemd this way.")

src/main.rs
flokli: See https://docs.rs/sd-notify/latest/src/sd_notify/lib.rs.html#105

You don't want to fail if informing about readyness failed, e.g. if the service is /not/ run via systemd and/or the communication with the unix socket not possible.

See also https://gist.github.com/grawity/6e5980981dccf66f554bbebb8cd169fc.

geofft:
I'm gonna merge this by itself and do an internal release to make sure nothing exciting happens. (The changes are de minimis so we don't need to bother with a CLA for this particular PR.)

Issue #37, opened 2022-10-10
provide host lookups


flokli:
Requires implementing at least:

 - `RequestType::GETAI`
 - `RequestType::GETHOSTBYADDR`
 - `RequestType::GETHOSTBYADDRv6`
 - `RequestType::GETHOSTBYNAME`
 - `RequestType::GETHOSTBYNAMEv6`

flokli:
@NinjaTrappeur and I have been taking a closer look at this and will send a PR once ready, just wanted to give a heads-up :-)

flokli:
A branch containing all of #38, #39 and #40 can be found at my [host-lookups](https://github.com/flokli/nsncd/tree/host-lookups) branch. We sent individual PRs so things can be reviewed individually.

There's various people testing this on their systems, see https://github.com/NixOS/nixpkgs/issues/135888#issuecomment-1278129060 for details.

flokli:
We combined #39 and #40 into #49, so they can be reviewed as a single unit.

leifwalsh:
Done in #71

PR #38, opened 2022-10-10
src/handlers.rs: explicitly handle INVALIDATE, SHUTDOWN, GETFD*

flokli:
In both cases, we want to no-op. The client doesn't expect any response from nscd, but let's at least log a debug message on why we're no-op'ing here, to distinguish from the operations that are not yet implemented.


flokli:
I saw the `GETFD*` requests on the wire when working on reverse host lookups (while adding `GETHOSTBYADDR*)` support as part of https://github.com/twosigma/nsncd/issues/37.

@geofft @leifwalsh WDYT?

leifwalsh:
Seems reasonable to me

geofft (approved):
Hmm, not unreasonable but I'm curious about the motivation. I feel like I'd do it the other way around - all of these really should get no response but the rest of them are unimplemented and it might be worth noting if we get one.

Should we (also? instead?) log the unimplemented requests? The enum derives Debug so we can just log the actual request value.

(Also now that I'm looking at this it occurs to me that `LASTREQ` isn't actually a valid protocol request. :) )

src/handlers.rs
```diff
@@ -111,21 +111,41 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
...
+            Ok(vec![])
+        }
+
+        // Not implemented (yet)
         RequestType::GETHOSTBYADDR
         | RequestType::GETHOSTBYADDRv6
         | RequestType::GETHOSTBYNAME
         | RequestType::GETHOSTBYNAMEv6
-        | RequestType::SHUTDOWN
         | RequestType::GETSTAT
```
geofft, line 145: This returns statistics about the daemon - you can also mark this as will-not-implement.

flokli: Why? I've been thinking about exposing some prometheus-style metrics on the number of requests, success rate and timing Information, by request type. That'd of course primarily be exposed via http, but if that getstat call is actually usable to expose some of these numbers, and there's some userspace tool to query and show, why not?

flokli:
> Hmm, not unreasonable but I'm curious about the motivation. I feel like I'd do it the other way around - all of these really should get no response but the rest of them are unimplemented and it might be worth noting if we get one.

I was mostly concerned about having a distinction in code if something has not been implemented yet, or if we consciously decided it's okay to return no data, and userspace can cope with it.

> 
> Should we (also? instead?) log the unimplemented requests? The enum derives Debug so we can just log the actual request value.

I think we already `debug!` the request in `main.rs`.

> (Also now that I'm looking at this it occurs to me that `LASTREQ` isn't actually a valid protocol request. :) )

Can you elaborate? It's not something that ever happens to be sent out by the client, or it's something where we should always pretend to have not yet have had any previous request?

flokli:
Btw, can these workflows be changed to not require approval? It'd be nice to immediately get feedback on the various GH action checks...

leifwalsh:
It's set to require approval for first time contributors, to prevent something like someone running a crypto miner using actions. Once we get you that CLA (#44) and merge some PRs from you you'll be good.

flokli:
Is this good to go?

leifwalsh:
> I feel like I'd do it the other way around - all of these really should get no response but the rest of them are unimplemented and it might be worth noting if we get one.
> 
> Should we (also? instead?) log the unimplemented requests? The enum derives Debug so we can just log the actual request value.

I like the gist of this: it clarifies which operations we'll never implement and makes it clear which remain to be implemented. We can log the unimplemented ones (at info? warn?) later.

I'm gonna merge this for now, we can tweak later (e.g. doing _something_ with `LASTREQ`).

PR #39, opened 2022-10-11
Implement the GETAI operation

NinjaTrappeur:
We implement the GETAI (aka. get address info) operation. Most of the glibc applications will rely on this operation to perform the hostname -> addrs resolutions.

We add some unit tests checking that we're correctly serializing the address info header and its payload. The expected data used in these tests has been extracted from various Nscd socket dumps. We also add an integration test checking that getaddrinfo is able to resolve localhost.

The Nix crate we rely on for the glibc FFI bindings does not wrap getaddrinfo yet. We had to pull the `dns-lookup` package to get those bindings.

Part of #37.

Note: on top of the automated tests, this PR has also been manually tested via a test VM.

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -14,17 +14,20 @@
  * limitations under the License.
  */
 
+use std::collections::HashSet;
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
 use std::os::unix::ffi::OsStrExt;
 
 use anyhow::{Context, Result};
 use atoi::atoi;
+use core::mem::size_of;
+use nix::sys::socket::{AddressFamily, IpAddr};
 use nix::unistd::{getgrouplist, Gid, Group, Uid, User};
 use slog::{debug, error, Logger};
 
 use super::protocol;
-use super::protocol::RequestType;
+use super::protocol::{AiResponse, AiResponseHeader, RequestType};
 
 /// Handle a request by performing the appropriate lookup and sending the
 /// serialized response back to the client.
@@ -111,6 +114,28 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
             };
             serialize_initgroups(groups)
         }
+        RequestType::GETAI => {
+            let hostname = CStr::from_bytes_with_nul(request.key)?.to_str()?;
+            let resp = dns_lookup::getaddrinfo(Some(hostname), None, None);
+
+            let ai_resp: AiResponse = match resp {
+                Ok(ai_resp_iter) => {
+                    let addrs: HashSet<IpAddr> = ai_resp_iter
+                        .filter_map(|e| e.ok())
+                        .map(|e| IpAddr::from_std(&e.sockaddr.ip()))
+                        .collect();
+                    AiResponse {
+                        canon_name: hostname.to_string(),
+                        addrs: addrs.iter().copied().collect::<Vec<IpAddr>>(),
+                    }
+                }
+                Err(_) => AiResponse {
+                    canon_name: hostname.to_string(),
+                    addrs: vec!()
+                },
+            };
+            Ok(serialize_address_info(&ai_resp)?)
+        }
         RequestType::GETHOSTBYADDR
         | RequestType::GETHOSTBYADDRv6
         | RequestType::GETHOSTBYNAME
@@ -121,7 +146,6 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
         | RequestType::GETFDPW
         | RequestType::GETFDGR
         | RequestType::GETFDHST
-        | RequestType::GETAI
         | RequestType::GETSERVBYNAME
         | RequestType::GETSERVBYPORT
         | RequestType::GETFDSERV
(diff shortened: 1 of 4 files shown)


src/handlers.rs
```diff
@@ -121,7 +126,19 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
         | RequestType::GETFDPW
         | RequestType::GETFDGR
         | RequestType::GETFDHST
-        | RequestType::GETAI
+        | RequestType::GETAI => {
+            debug!(log, "handling GETAI"; "request" => ?request);
+            let request_str = CStr::from_bytes_with_nul(request.key)?.to_str()?;
+
+            // The original Nscd GETAI implementation can only fail in
```
flokli, line 133:
```suggestion
            // The original nscd GETAI implementation can only fail in
```

src/handlers.rs
```diff
@@ -279,4 +394,110 @@ mod test {
...
+    #[test]
+    fn test_handle_request_getai () {
+        let request = protocol::Request {
+            ty: protocol::RequestType::GETAI,
+            key: &CString::new("localhost".to_string())
+                .unwrap()
+                .into_bytes_with_nul(),
+        };
+
+        // The getaddrinfo addresses ordering is quite unprediteable.
```
flokli, line 407:
```suggestion
        // The getaddrinfo addresses ordering is quite unpredictable.
```

src/handlers.rs
```diff
@@ -279,4 +394,110 @@ mod test {
...
+
+        let expected = protocol::AI_RESPONSE_HEADER_NOT_FOUND.as_slice();
+        let output =
+            handle_request(&test_logger(), &request).expect("should handle request with no error");
+
+        assert_eq!(expected, output);
+    }
+
+    /// This test compares the [serialize_adress_info] output to a
+    /// byte array generated by the original Nscd implementation.
```
flokli, line 457:
```suggestion
    /// byte array generated by the original nscd implementation.
```

src/handlers.rs
```diff
@@ -279,4 +394,110 @@ mod test {
...
+    #[test]
+    fn test_serialize_address_info() {
+        let test_response = AiResponse {
+            addrs: vec!(
+                IpAddr::new_v6(0,0,0,0,0,0,0,1),
+                IpAddr::new_v4(127,0,0,1)
```
flokli, line 463: did this not get reformatted?

src/handlers.rs
```diff
@@ -279,4 +394,102 @@ mod test {
...
+                IpAddr::new_v6(0, 0, 0, 0, 0, 0, 0, 1),
+                IpAddr::new_v4(127, 0, 0, 1),
+            ],
+            canon_name: "localhost".to_string(),
+        };
+
+        let expected: Vec<u8> = vec![
+            // Response Header
+            // ===============
+            // version number = 2
```
flokli, line 474: The rustfmt did fold the comments up into the wrong line.

NinjaTrappeur: Aha, thanks!

src/handlers.rs
```diff
@@ -279,4 +396,101 @@ mod test {
...
+        let request = protocol::Request {
+            ty: protocol::RequestType::GETAI,
+            key: &CString::new("localhost".to_string())
+                .unwrap()
+                .into_bytes_with_nul(),
+        };
+
+        // The getaddrinfo addresses ordering is quite unpredictable.
+        // Trying the two potential orderings.
+        // We could also be on a IPv4-only machine on CI.
```
flokli, line 411: Is this because `/etc/hosts` doesn't contain the entry, or because the loopback interface in CI doesn't contain a v6 address?

I fear this might also be a problem on other machines running `cargo test` - we should make this a bit more reliable…

NinjaTrappeur: > Is this because /etc/hosts doesn't contain the entry, or because the loopback interface in CI doesn't contain a v6 address?

It's an issue on kernels configured not to support IPv6.

> I fear this might also be a problem on other machines running cargo test - we should make this a bit more reliable…

Yeah, I'm not super happy with that either.

The idea here is not only to test the serialized output itself, but also the content of the response. What we're sending to the wire is not exactly what `getaddrinfo` returns. There's some mapping logic happening in the `handlers::getaddrinfo` function, that's partially what we're testing here.

I put some thoughts on that and did not find a more robust way to check this function output than resolving `localhost` :( 

Do you have any idea?

flokli: I think:
 - we should test the `serialize_` with manually crafted input
 - test a `GETAI` request for `localhost` to not fail (but don't peek into the response)
 - … and expect `dns_lookup::getaddrinfo` to do the right thing.

NinjaTrappeur: > we should test the serialize_ with manually crafted input

This is already tested by https://github.com/twosigma/nsncd/pull/39/files#diff-148e623ad4cdaab49b81901460099f4efe7900e8c57e4f7a0ff6778589be01f8R464

> test a GETAI request for localhost to not fail (but don't peek into the response) … and expect dns_lookup::getaddrinfo to do the right thing.

This test is not really testing `dns_lookup::getaddrinfo` but rather than `handlers::getaddrinfo` properly massage its result to our needs. Removing the duplicates entries for now. getaddrinfo returns one item per `AddrInfo::socktype`, we only care about the IP adresses, we want to filter out the duplicates.

flokli: WDYT about https://github.com/flokli/nsncd/commit/d7f89f81cdb6e988fdf42b2479907e65b3ca459c?

I still don't know how to handle the fact that localhost might only return v4 addresses, depending on the host configured, but at least now this tests the deduplication again.

NinjaTrappeur: sounds good

leifwalsh:
@NinjaTrappeur can you describe your manual test? I wonder if that's something we can get working in Actions.

flokli:
> can you describe your manual test? I wonder if that's something we can get working in Actions.

We use a NixOS VM test, using a custom version of getent. This test makes use of the fact that external glibc modules only work via nscd protocol on NixOS, and uses the nss-systemd module too. I'm not sure this will be something trivial to port to a container.

We could add the VM test code to this repository, and tweak it so that it uses the nsncd code from here, but running it in GH actions might not work (nested virt).

See https://github.com/flokli/nixpkgs/blob/e26519527879774b6a2551b1704598039cdbcb6e/nixos/tests/nscd.nix for the test code.

geofft:
On the actions front - you can reproduce the issue with Nix (nixpkgs) on a non-NixOS OS (in fact that's our use case). There is a host-wide libc, but Nix-built software doesn't use it. So I don't think we need NixOS specifically; we can just install Nix inside the actions VM running Ubuntu (or whatever) and make sure a `getent` call from nixpkgs.glibc works. That's probably worth doing anyway, since that's a more end-to-end test than our current one (e.g., it will catch if the NSCD protocol changes in some unfortunate way). I can give that a shot.

geofft:
Both this and #40 seem reasonable.

Note that they both have to be merged in the same release because glibc's NSCD client disables lookups for the whole hosts family of operations if one operation comes back unimplemented (see also #26, where it disabled lookups for the whole groups family of operations), but that shouldn't be hard to do. I don't see any other host operations I think.

I'd like to add an end-to-end test and we should sort out the CLA stuff (#44) but I'm actually curious to run this internally on my dev box (i.e. without yet writing config to let you disable this)  and see how it holds up. I also want to stare a little bit harder at some of the `_nss_gethostbyname4` stuff that isn't super well-documented to make sure we're indeed calling the right NSS functions, but I think this code does (We can test that in the end-to-end test.)

Thank you for the contributions!

PR #40, opened 2022-10-13
implement GETHOSTBY{ADDR,NAME}[v6]

flokli:
This adds support for the following request types:

 - GETHOSTBYADDR
 - GETHOSTBYADDRv6
 - GETHOSTBYNAME
 - GETHOSTBYNAMEv6

Part of #37.

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -16,11 +16,13 @@
 
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
-use std::net::IpAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::os::unix::ffi::OsStrExt;
 
 use anyhow::{bail, Context, Result};
 use atoi::atoi;
+use dns_lookup::getnameinfo;
+use nix::libc::NI_NUMERICSERV;
 use nix::unistd::{getgrouplist, Gid, Group, Uid, User};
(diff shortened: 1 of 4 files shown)


src/handlers.rs
```diff
@@ -111,11 +114,106 @@ pub fn handle_request(log: &Logger, request: &protocol::Request) -> Result<Vec<u
...
+            let hostname = CStr::from_bytes_with_nul(request.key)?.to_str()?;
+            let hints = AddrInfoHints {
+                socktype: SOCK_STREAM,
+                ..AddrInfoHints::default()
+            };
+
+            let host = match getaddrinfo(Some(hostname), None, Some(hints)) {
+                Ok(addrs) => {
+                    let addresses: std::io::Result<Vec<_>> =
+                        addrs.map(|r| r.map(|a| a.sockaddr.ip())).collect();
```
flokli, line 175: Here (and in the v6 case further down) I forgot to filter, and in case both address families are returned, `serialize_hosts` bails out silently with an error (because we have no logger there).

I'll send a fix for this in a bit.

flokli: Done!

flokli:
Closing in favor of https://github.com/twosigma/nsncd/pull/49.

PR #41, opened 2022-10-14
Build in Debian 10 container

geofft:
This ensures that we're compatible with Debian's packaged version of rustc. Ideally we would run the full CI workflow (package build and installation) on Debian 10, but since that's best done on a full system and not a container to test systemd packaging etc., leave it alone for now.

### DIFF
--- a/.github/workflows/rust-debian10.yml
+++ b/.github/workflows/rust-debian10.yml
@@ -1,0 +1,26 @@
+name: "Debian 10"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-debian10.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    runs-on: ubuntu-latest
+    container: debian:10
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install dependencies
+      run: |
+        apt-get update
+        apt-get install build-essential dpkg-dev
```
leifwalsh, line 23: I think you want `-y` and maybe `--no-install-recommends`

geofft:
OK, [CI is failing the way I expect now](https://github.com/twosigma/nsncd/actions/runs/3256951687/jobs/5347888902) and will work once #45 lands.

leifwalsh (approved):

Issue #42, opened 2022-10-15
runtime configuration


leifwalsh:
@flokli and @ninjatrappeur have been looking at adding support for more operations in #37.

The original purpose for, and our internal usage of, nsncd only needed user and group lookups to be served by nsncd, and while we're comfortable with the current performance and backpressure mechanisms, we have had some scares before where nsncd got overwhelmed by requests (hence #27 #28 #33 etc.). We're a little wary of adding more supported operations to our build.

I think we'd feel more comfortable if there were a way to configure which operations nsncd will handle, and all of us might appreciate runtime configuration anyway, to control the supported operations, threadpool settings, #15, etc.

I haven't looked at Rust configuration libraries but something that handles a common format like TOML and maybe supports environment variable overrides should be out there and sounds nice to me.

leifwalsh:
https://docs.rs/config/latest/config/ and https://docs.rs/figment/latest/figment/ look promising at first glance.

blinsay:
Clap also has support for env variables these days.

https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.env

PR #49, opened 2022-10-20
Implement host lookups 

flokli:
This adds support for the following request types:

 - GETAI
 - GETHOSTBYADDR
 - GETHOSTBYADDRv6
 - GETHOSTBYNAME
 - GETHOSTBYNAMEv6

Cherry-picked into one PR due to https://github.com/twosigma/nsncd/pull/39#issuecomment-1279858396:

> Note that they both have to be merged in the same release because glibc's NSCD client disables lookups for the whole hosts family of operations if one operation comes back unimplemented (see also https://github.com/twosigma/nsncd/pull/26, where it disabled lookups for the whole groups family of operations), but that shouldn't be hard to do. I don't see any other host operations I think.

Fixes #37.
Supersedes #39 and #40.

PR #51, opened 2022-11-19
add simple runtime configuration

leifwalsh:
This allows us to disable any request types we don't want nsncd to handle.

fixes runtime configuration #42

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
 # 3.2 MSRV is too new: rust edition 2021
-clap = { version = "~3.1", features = ["derive", "env"] }
+clap = { version = "=3.1.12", features = ["derive", "env"] }
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
-clap = { version = "^4.0.26", features = ["derive", "env"] }
+clap = { version = "^3.2", features = ["derive", "env"] }
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,6 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
-# version 3 MSRV is too new: pulls in os_str_bytes 6.0 which has too high an MSRV
-clap = "2"
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ once_cell = "~1.14"
 
 [dev-dependencies]
 criterion = "^0.3"
+temp-env = "^0.3"
 
 [[bench]]
 name = "user"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
+clap = { version = "^4.0.26", features = ["derive", "env"] }
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,8 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
-clap = { version = "^3.2", features = ["derive", "env"] }
+# 3.2 MSRV is too new: rust edition 2021
+clap = { version = "~3.1", features = ["derive", "env"] }
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
-# 3.2 MSRV is too new: rust edition 2021
-clap = { version = "=3.1.12", features = ["derive", "env"] }
+# version 3 MSRV is too new: pulls in os_str_bytes 6.0 which has too high an MSRV
+clap = "2"
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
+static_assertions = "1.1.0"
 
 [dev-dependencies]
 criterion = "^0.3"
--- a/src/config.rs
+++ b/src/config.rs
@@ -63,6 +63,31 @@ const OPS_BY_DATABASE: &[(&str, &[RequestType])] = &[
 ];
 
 impl Config {
+    /// Parse config out of the environment.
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -28,6 +28,7 @@ num-traits = "^0.2"
 sd-notify = "~0.3"
 # Indirect dependency, constrained for MSRV
 once_cell = "~1.14"
+clap = { version = "^4.0.26", features = ["derive", "env"] }
```
leifwalsh, line 31: I'm a little worried about what this does to MSRV, hopefully CI will catch it.

src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+use clap::builder::BoolishValueParser;
+
+use super::protocol::RequestType;
+
+pub trait RequestTypeIgnorer {
+    fn should_ignore(&self, ty: &RequestType) -> bool;
+}
+
+#[derive(clap::Parser, Clone, Copy, Debug, Default)]
+pub struct Config {
```
leifwalsh, line 30: Someone with more macro skills than me could probably reduce the boilerplate here.

Cargo.toml
leifwalsh: Ok yeah I think we need clap 3 which means I need to figure the api out again.

leifwalsh:
Ok, I tried downgrading all the way to clap v2, whose API is miserable compared to modern clap, and it still doesn't compile on debian 10. I think I'm giving up on this path for now.

src/config.rs
blinsay: We absolutely should do something less boiler-plate-y. I think it makes sense to keep a set of enabled/disabled request types.

```rust
impl FromStr for RequestType {
  ... 
}

impl Config {
  fn from_env() -> Result<Self> {
    let mut disabled: HashSet<RequestType> = ...'
    for (key, val) in std::env::vars() {
      if key.starts_with("NSNCD_IGNORE") && env_bool(val) {
         let type: RequestType = string.strip_prefix().parse().expect("this is psuedocode fix it");
         disabled.insert(type);
      }
    }
    return disabled;
  }
}

```

This way we could also decide to define groups of request types to enable/disable together as well. I'd defer to @geofft on which groups make sense (and we can also bechmark and figure out whether a HashSet or just a slice is faster for lookups once we're done).

Unless there's a crate you already know about that has a nice derive-macro for FromStr I will totally help you write some `macro_rules` nonsense for `RequestType` boilerplate.

src/config.rs
```diff
@@ -0,0 +1,335 @@
...
+            ignore_getservbyport: Default::default(),
+            ignore_getfdserv: Default::default(),
+            ignore_getfdnetgr: Default::default(),
+            ignore_getnetgrent: Default::default(),
+            ignore_innetgr: Default::default(),
+        }
+    }
+}
+
+fn env_positive_usize(var: &'static str, default: usize) -> Result<usize> {
```
blinsay, line 147: nit: I generally prefer to flatten out something like this instead of an `if` nested inside an `if-let`. if it weren't for MSRV i'd suggest `guard-let`.

```rust
let s = match env::var(var) {
  Ok(s) => s,
  Err(_) => return default,
};

let int_value: usize = s.parse().with_context(|| "good error message")?;
if val > 0 {
  Ok(val)
} else {
   Err("better error message")
}
```

I would also be down for something like this, but having the default be a string always feels a little silly.

```rust
let int_val: usize = env::var(var).unwrap_or(default).parse().with_context(...)?;
if val > 0 {
  Ok(val)
} else {
   Err("better error message")
}
```

blinsay (changes requested):

src/config.rs
leifwalsh: Yeah, I thought about something like HashSet, but I figured a match and an index into a struct was simplest and fastest. I think this gets us off the ground in terms of having control, and we can work on the ergonomics later.

I can see an argument for operation groups like `--ignore-users` or `--ignore-groups` or `--ignore-network`. 

There are crates out there that let you derive iterators for enums and maybe you could use those in a macro, but I got shook by MSRV tbqh.

blinsay: > Yeah, I thought about something like HashSet, but I figured a match and an index into a struct was simplest and fastest. I think this gets us off the ground in terms of having control, and we can work on the ergonomics later.

I think the HashSet thing is much easier to maintain than the current thing, but I'll defer to you unless we get a tiebreaker vote. I'm not really worried about it taking the extra day or two to get done.

The more I think about it, the more I think it might be easier to do groups than individual messages because there's less config to parse.

```rust
if std::env::var("NSNCD_ENABLE_GROUPS").is_ok() {
   enabled.extend(GROUPS_MESSAGES)
}

if std::env::var("NSNCD_ENABLED_NETWORK").is_ok() {
  enabled.extend(ADDRESS_MESSAGES);
}
```

If you do want to keep going down the road enabling/disabling individual request types, here's a macro [(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=71672b0d8120b765e83e453b6b353062) to do `FromStr` for `RequestType`. We'd have to do a tiny bit of shenanigans when parsing the env variables because `stringify!` preserves case and case conversion isn't const.

```rust
macro_rules! request_type {
    ($($name:ident),*) => {
        #[derive(Debug)]
        pub enum RequestType {
            $(
            $name
            ),*
        }
        
        impl std::str::FromStr for RequestType {
            type Err = ();
            
            fn from_str(s: &str) -> Result<RequestType, ()> {
                match s {
                    $(
                     stringify!($name) => Ok(RequestType::$name),
                    )*
                    _ => Err(()),
                }
            }
        }
    }
}

request_type! {
    Foo,
    Bar,
    Baz
}

fn main() {
    let v: Result<RequestType, _> = "Foo".parse();
    
    println!("{v:?}");
}
```

flokli: I'd personally prefer "groups of request types" that can be split, rather than individual requests.

As we could see in https://github.com/twosigma/nsncd/pull/26 and https://github.com/twosigma/nsncd/pull/49, there's often multiple request types that need to be handled to get a specific lookup to work.

Providing fine-grained levers that then cause more to be disabled, because that's how client-side detection logic works seems to be very confusing.

I guess you want this mostly to be able to disable host lookups, in case this is causing issues? I'd probably really limit it to only that, because we know having these disabled might be a valid "working" configuration.

leifwalsh: Yeah, thinking better about this the controls should mirror nsswitch.conf. Lemme take this back to the drawing board.

src/protocol.rs
```diff
@@ -37,7 +37,7 @@ pub const VERSION: i32 = 2;
 
 /// Available services. This enum describes all service types the nscd protocol
 /// knows about, though we only implement `GETPW*`, `GETGR*`, and `INITGROUPS`.
-#[derive(Debug, FromPrimitive)]
+#[derive(Clone, Copy, Debug, FromPrimitive, ToPrimitive)]
```
blinsay, line 40: Can we do some kind of static assert on the max value of `ToPrimitive` to make sure that we have a large enough set allocated in Config?

src/config.rs
blinsay: Rad, I like this approach a lot more.

src/config.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// - NSNCD_IGNORE_SERVICES
+    ///
+    /// These variables must be either `true` or `false`. The default is
+    /// `false` (don't ignore any requests). If one of these variables is set
+    /// to true, `nsncd` will not respond to the requests related to that
+    /// database.
+    ///
+    /// Some request types may be ignored by the implementation (e.g. the ones
+    /// that request a file descriptor pointing into internal cache
+    /// structures).
```
blinsay, line 90: Let's add the config env vars to the README as well.

src/protocol.rs
leifwalsh: I added something, thanks. I think I'd feel better if I checked all values of `RequestType` but I think this is good enough for now.

src/config.rs
leifwalsh: done

blinsay (approved):

PR #59, opened 2023-04-14
build debian package as a separate job and upload artifacts

leifwalsh:


leifwalsh:
This created https://github.com/twosigma/nsncd/releases/tag/untagged-b26dbd973fffb1784b77

Still need to fiddle with making it do this just for tagged releases, and we should talk about when/how we want to create tags and manage the changelog.

leifwalsh:
I think I want to get it working with the existing debian packaging stuff first, and then try to switch to cargo deb. For one thing, cargo deb requires a newer rust than we're currently on. I'd like to introduce just one change at a time.

PR #62, opened 2023-04-29
Revert "debian: Explicitly build deb with xz"

geofft:
This reverts commit 2366e0124d0f03bb7cf3d7cd03053208ad5f65cb.

PR #63, opened 2023-05-02
Revert "debian: Explicitly build deb with xz"

leifwalsh:
replaces #62 since #61 erased part of the commit it was reverting


geofft (approved):
OK with me but we should pick a version number for releasing #61 (reusing 1.4.1 should be fine)

PR #71, opened 2023-10-25
implement host lookup support

flokli:
This adds support for the GETAI, GETHOSTBYADDR, GETHOSTBYADDRv6, GETHOSTBYNAME, GETHOSTBYNAMEv6 request types.

For the more complex GETAI lookup, we use the dns_lookup crate.

In previous iterations of this change, we also used the same underlying getaddrinfo call to respond to the  gethostbyname/gethostbyaddr operations.

Even though gethostbyname/gethostbyaddr officially are deprecated, there's a lot of tools still using it, and relying on them behaving differently.

So it's important to still implement it, with exactly the same behaviour, to prevent some tools from breaking (like `hostname --fqdn`, see https://github.com/nix-community/nsncd/issues/4).

We FFI the right Glibc gethostbyname_2r/gethostbyaddr_2r (now deprecated) functions and use it to back the GETHOSTBYNAME, GETHOSTBYNAME6, GETHOSTBYADDR, and GETHOSTBYADDR6 Nscd interfaces.

Took us three tries to get this right. This is actually the third full rewrite.

The Nscd behaviour for these two legacy functions is *really* confusing. We're supposed to ignore the herrno (herrno != errno!!) and set it to 0 if gethostbyaddr/name returns a non-null hostent. If we end up with a null hostent, we return the herrno together with a dummy hostent header.

I tried to keep things as safe as possible by extracting the glibc hostent to a proper rust structure. This structure mirrors the libc hostent in a Rust idiomatic way. We should probably try to upstream the FFI part of this commit to the Nix crate at some point.

Fixes https://github.com/nix-community/nsncd/issues/4

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -171,6 +171,7 @@ fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
         aliases,
         addr_type: value.h_addrtype,
         addr_list,
+        // If we're here, glibc gave us an hostent. We should discard herrno to match the nscd behaviour.
         herrno: 0,
     })
 }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-use anyhow::bail;
+use anyhow::anyhow;
 use nix::libc::{self};
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
@@ -99,7 +99,7 @@ impl Hostent {
     /// convey a lookup error.
     /// NOTE: herrno is different from errno.h.
     /// This is a glibc quirk, I have nothing to do with that, don't blame me :)
-    fn error_value(herrno: i32) -> Self {
+    pub fn error_value(herrno: i32) -> Self {
         // This is a default hostent header
         Hostent {
             name: CString::default(),
@@ -111,25 +111,38 @@ impl Hostent {
     }
 }
 
-fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
+/// Structure used to represent a gethostbyxxx error.
+///
+/// These operations can fail in two major ways: either they'll fail
+/// returning a Hostent, in which case they return a HError code that
+/// should be returned to the glibc client together with a dummy error
+/// Hostent. Either as an "internal failure". In that case, we won't
+/// be able to return anything to the GLibc client.
+#[derive(Debug)]
+pub enum HostentError {
+    HError(i32),
+    Other(anyhow::Error)
+}
+
+fn from_libc_hostent(value: libc::hostent) -> Result<Hostent, HostentError> {
     // validate value.h_addtype, and bail out if it's unsupported
     if value.h_addrtype != libc::AF_INET && value.h_addrtype != libc::AF_INET6 {
-        bail!("unsupported address type: {}", value.h_addrtype);
+        return Err(HostentError::Other(anyhow!("unsupported address type: {}", value.h_addrtype)));
     }
 
     // ensure value.h_length matches what we know from this address family
     if value.h_addrtype == libc::AF_INET && value.h_length != 4 {
-        bail!("unsupported h_length for AF_INET: {}", value.h_length);
+        return Err(HostentError::Other(anyhow!("unsupported h_length for AF_INET: {}", value.h_length)));
     }
     if value.h_addrtype == libc::AF_INET6 && value.h_length != 16 {
-        bail!("unsupported h_length for AF_INET6: {}", value.h_length);
+        return Err(HostentError::Other(anyhow!("unsupported h_length for AF_INET6: {}", value.h_length)));
     }
 
     // construct the name field.
     // Be careful about null pointers or invalid utf-8 strings, even though this
     // shouldn't happen.
     if value.h_name.is_null() {
-        bail!("h_name is null");
+        return Err(HostentError::Other(anyhow!("h_name is null")));
     }
     let name = unsafe { CStr::from_ptr(value.h_name) };
 
@@ -182,16 +195,16 @@ fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
 fn unmarshal_gethostbyxx(
     hostent: *mut libc::hostent,
     herrno: libc::c_int,
-) -> anyhow::Result<Hostent> {
+) -> Result<Hostent, HostentError> {
     if !hostent.is_null() {
         let res = from_libc_hostent(unsafe { *hostent })?;
         Ok(res)
     } else {
-        Ok(Hostent::error_value(herrno))
+        Err(HostentError::HError(herrno))
     }
 }
 
-pub fn gethostbyaddr_r(addr: LibcIp) -> anyhow::Result<Hostent> {
+pub fn gethostbyaddr_r(addr: LibcIp) -> Result<Hostent, HostentError> {
     let (addr, len, af) = match addr {
         LibcIp::V4(ref ipv4) => (ipv4 as &[u8], 4, libc::AF_INET),
         LibcIp::V6(ref ipv6) => (ipv6 as &[u8], 16, libc::AF_INET6),
@@ -235,7 +248,7 @@ pub fn gethostbyaddr_r(addr: LibcIp) -> anyhow::Result<Hostent> {
 /// Typesafe wrapper around the gethostbyname2_r glibc function
 ///
 /// af is either nix::libc::AF_INET or nix::libc::AF_INET6
-pub fn gethostbyname2_r(name: String, af: libc::c_int) -> anyhow::Result<Hostent> {
+pub fn gethostbyname2_r(name: String, af: libc::c_int) -> Result<Hostent, HostentError> {
     let name = CString::new(name).unwrap();
 
     // Prepare a libc::hostent and the pointer to the result list,
@@ -278,12 +291,11 @@ pub fn gethostbyname2_r(name: String, af: libc::c_int) -> anyhow::Result<Hostent
 fn test_gethostbyname2_r() {
     disable_internal_nscd();
 
-    let result: Result<Hostent, anyhow::Error> =
+    let result =
         gethostbyname2_r("localhost.".to_string(), libc::AF_INET);
-
     result.expect("Should resolve IPv4 localhost.");
 
-    let result: Result<Hostent, anyhow::Error> =
+    let result =
         gethostbyname2_r("localhost.".to_string(), libc::AF_INET6);
     result.expect("Should resolve IPv6 localhost.");
 }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-use anyhow::bail;
+use anyhow::{anyhow, bail};
 use nix::libc::{self};
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
@@ -94,6 +94,23 @@ pub struct Hostent {
     pub herrno: i32,
 }
 
+impl Hostent {
+    /// Given a herrno, constructs the hostent header we're supposed to use to
+    /// convey a lookup error.
+    /// NOTE: herrno is different from errno.h.
+    /// This is a glibc quirk, I have nothing to do with that, don't blame me :)
+    fn error_value(herrno: i32) -> Self {
+        // This is a default hostent header
+        Hostent {
+            name: "".to_string(),
+            aliases: Vec::new(),
+            addr_type: -1,
+            addr_list: Vec::new(),
+            herrno,
+        }
+    }
+}
+
 fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
     // validate value.h_addtype, and bail out if it's unsupported
     if value.h_addrtype != libc::AF_INET && value.h_addrtype != libc::AF_INET6 {
@@ -108,47 +125,53 @@ fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
         bail!("unsupported h_length for AF_INET6: {}", value.h_length);
     }
 
-    let name = unsafe { CStr::from_ptr(value.h_name).to_str().unwrap().to_string() };
+    // construct the name field.
+    // Be careful about null pointers or invalid utf-8 strings, even though this
+    // shouldn't happen.
+    if value.h_name.is_null() {
+        bail!("h_name is null");
+    }
+    let c_name = unsafe { CStr::from_ptr(value.h_name) };
+    let name = c_name
+        .to_str()
+        .map_err(|_| anyhow!("unable to convert {:?} to string", c_name))?
+        .to_string();
 
-    Ok({
-        // construct the list of aliases. keep adding to value.h_aliases until we encounter a null pointer.
-        let mut aliases: Vec<String> = Vec::new();
-        let mut h_alias_ptr = value.h_aliases as *const *const libc::c_char;
-        while !(unsafe { *h_alias_ptr }).is_null() {
-            aliases.push(unsafe { CStr::from_ptr(*h_alias_ptr).to_str().unwrap().to_string() });
-            // increment
-            unsafe {
-                h_alias_ptr = h_alias_ptr.add(1);
-            }
+    // construct the list of aliases. keep adding to value.h_aliases until we encounter a null pointer.
+    let mut aliases: Vec<String> = Vec::new();
+    let mut h_alias_ptr = value.h_aliases as *const *const libc::c_char;
+    while !(unsafe { *h_alias_ptr }).is_null() {
+        aliases.push(unsafe { CStr::from_ptr(*h_alias_ptr).to_str().unwrap().to_string() });
+        // increment
+        unsafe {
+            h_alias_ptr = h_alias_ptr.add(1);
         }
-        // value.h_addrtype
+    }
 
-        // construct the list of addresses.
-        let mut addr_list: Vec<std::net::IpAddr> = Vec::new();
+    // construct the list of addresses.
+    let mut addr_list: Vec<std::net::IpAddr> = Vec::new();
 
-        // copy the pointer into a private variable that we can mutate
-        // h_addr_list is a pointer to a list of pointers to addresses.
-        // h_addr_list[0] => ptr to first address
-        // h_addr_list[1] => null pointer (end of list)
-        let mut h_addr_ptr = value.h_addr_list as *const *const libc::c_void;
-        while !(unsafe { *h_addr_ptr }).is_null() {
-            if value.h_addrtype == libc::AF_INET {
-                let octets: [u8; 4] = unsafe { std::ptr::read((*h_addr_ptr) as *const [u8; 4]) };
-                addr_list.push(std::net::IpAddr::V4(std::net::Ipv4Addr::from(octets)));
-            } else {
-                let octets: [u8; 16] = unsafe { std::ptr::read((*h_addr_ptr) as *const [u8; 16]) };
-                addr_list.push(std::net::IpAddr::V6(std::net::Ipv6Addr::from(octets)));
-            }
-            unsafe { h_addr_ptr = h_addr_ptr.add(1) };
+    // [value.h_addr_list] is a pointer to a list of pointers to addresses.
+    // h_addr_list[0] => ptr to first address
+    // h_addr_list[n] => null pointer (end of list)
+    let mut h_addr_ptr = value.h_addr_list as *const *const libc::c_void;
+    while !(unsafe { *h_addr_ptr }).is_null() {
+        if value.h_addrtype == libc::AF_INET {
+            let octets: [u8; 4] = unsafe { std::ptr::read((*h_addr_ptr) as *const [u8; 4]) };
+            addr_list.push(std::net::IpAddr::V4(std::net::Ipv4Addr::from(octets)));
+        } else {
+            let octets: [u8; 16] = unsafe { std::ptr::read((*h_addr_ptr) as *const [u8; 16]) };
+            addr_list.push(std::net::IpAddr::V6(std::net::Ipv6Addr::from(octets)));
         }
+        unsafe { h_addr_ptr = h_addr_ptr.add(1) };
+    }
 
-        Hostent {
-            name,
-            aliases,
-            addr_type: value.h_addrtype,
-            addr_list,
-            herrno: 0,
-        }
+    Ok(Hostent {
+        name,
+        aliases,
+        addr_type: value.h_addrtype,
+        addr_list,
+        herrno: 0,
     })
 }
 
@@ -164,21 +187,10 @@ fn unmarshal_gethostbyxx(
     herrno: libc::c_int,
 ) -> anyhow::Result<Hostent> {
     if !hostent.is_null() {
-        let res = from_libc_hostent(unsafe { *hostent } )?;
+        let res = from_libc_hostent(unsafe { *hostent })?;
         Ok(res)
     } else {
-        Ok(
-            // This is a default hostent header we're supposed to use
-            // to convey a lookup error. This is a glibc quirk, I have
-            // nothing to do with that, don't blame me :)
-            Hostent {
-                name: "".to_string(),
-                aliases: Vec::new(),
-                addr_type: -1,
-                addr_list: Vec::new(),
-                herrno,
-            },
-        )
+        Ok(Hostent::error_value(herrno))
     }
 }
 
@@ -214,7 +226,7 @@ pub fn gethostbyaddr_r(addr: LibcIp) -> anyhow::Result<Hostent> {
             )
         };
 
-        if ret == libc::ERANGE && buf.capacity() < 10 * 1000 * 1000  {
+        if ret == libc::ERANGE && buf.capacity() < 10 * 1000 * 1000 {
             buf.reserve(buf.capacity() * 2);
         } else {
             break;
@@ -261,7 +273,7 @@ pub fn gethostbyname2_r(name: String, af: libc::c_int) -> anyhow::Result<Hostent
         } else {
             break;
         }
-    };
+    }
     unmarshal_gethostbyxx(hostent_result, herrno)
 }
 
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-use anyhow::{anyhow, bail};
+use anyhow::bail;
 use nix::libc::{self};
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
@@ -87,8 +87,8 @@ mod glibcffi {
 /// TryFrom `libc:hostent` trait.
 #[derive(Clone, Debug)]
 pub struct Hostent {
-    pub name: String,
-    pub aliases: Vec<String>,
+    pub name: CString,
+    pub aliases: Vec<CString>,
     pub addr_type: i32,
     pub addr_list: Vec<std::net::IpAddr>,
     pub herrno: i32,
@@ -102,7 +102,7 @@ impl Hostent {
     fn error_value(herrno: i32) -> Self {
         // This is a default hostent header
         Hostent {
-            name: "".to_string(),
+            name: CString::default(),
             aliases: Vec::new(),
             addr_type: -1,
             addr_list: Vec::new(),
@@ -131,17 +131,13 @@ fn from_libc_hostent(value: libc::hostent) -> anyhow::Result<Hostent> {
     if value.h_name.is_null() {
         bail!("h_name is null");
     }
-    let c_name = unsafe { CStr::from_ptr(value.h_name) };
-    let name = c_name
-        .to_str()
-        .map_err(|_| anyhow!("unable to convert {:?} to string", c_name))?
-        .to_string();
+    let name = unsafe { CStr::from_ptr(value.h_name) };
 
     // construct the list of aliases. keep adding to value.h_aliases until we encounter a null pointer.
-    let mut aliases: Vec<String> = Vec::new();
+    let mut aliases: Vec<CString> = Vec::new();
     let mut h_alias_ptr = value.h_aliases as *const *const libc::c_char;
     while !(unsafe { *h_alias_ptr }).is_null() {
-        aliases.push(unsafe { CStr::from_ptr(*h_alias_ptr).to_str().unwrap().to_string() });
+        aliases.push(unsafe { CStr::from_ptr(*h_alias_ptr).to_owned() });
         // increment
         unsafe {
             h_alias_ptr = h_alias_ptr.add(1);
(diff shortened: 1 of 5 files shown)


flokli:
We tested these changes extensively using [sockburp](https://github.com/flokli/sockburp/wiki/glibc%E2%80%90nscd---nsncd-on-NixOS) on the [host-lookups](https://github.com/nix-community/nsncd/tree/host-lookups) branch on the `nix-community` fork. This is a cherry-pick of the latest iteration.

We'll give it some more testing on our side, but wanted to send this out already, so you can take a look.

flokli:
Marking as ready for review, PTAL :-)

picnoir:
There's something still off with the getaddrinfo canonical name.

flokli:
The remaining issue is fixed too, and cherry-picked here - see https://github.com/nix-community/nsncd/pull/10 for details.

src/handlers.rs
```diff
@@ -134,6 +142,125 @@ pub fn handle_request(
...
 
+        RequestType::GETAI => {
+            let hostname = CStr::from_bytes_with_nul(request.key)?.to_str()?;
+            // Boths hints are necessary to mimick the glibc behaviour.
+            let hints = AddrInfoHints {
+                // The canonical name will be filled in the first
+                // addrinfo struct returned by getaddrinfo.
+                flags: AI_CANONNAME,
+                // Deduplicate the resulting addrs.
+                socktype: SOCK_STREAM,
```
leifwalsh, line 153: Why `SOCK_STREAM` here? I guess, what if the nscd client is trying to do `SOCK_DGRAM`, what do we do and what would real `nscd` do?

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+/// This wire format has been implemented by reading the `addhstaiX`
+/// function living in the `nscd/aicache.c` glibc file. We copy the
+/// exact same behaviour, aside from the caching part.
+///
+/// The wire getaddrinfo call result is serialized like this:
+///
+/// 1. version: int32. Hardcoded to 2.
+/// 2. found: int32. 1 if we have a result, 0 if we don't.
+/// 3. naddrs: int32. Number of IPv4/6 adresses we're about to write.
+/// 4. addrslen: int32. Total lenght of the IPv4/6 adresses we're
```
leifwalsh, line 508:
```suggestion
/// 4. addrslen: int32. Total length of the IPv4/6 adresses we're
```

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+///    about to write.
+/// 5. canonlen: int32. Total lenght of the null-terminated canonical
```
leifwalsh, line 510:
```suggestion
/// 5. canonlen: int32. Total length of the null-terminated canonical
```

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+///    name string.
+/// 6. error: int32. Error code. Always 0 in the current nscd
+///           implementation.
+/// 7. addrs: \[BE-encoded IPv4/IPv6\]. We sequentially write the
+///    IPv4 and IPv6 bytes using a big endian encoding. There's no
+///    padding, an IPv4 will be 4 bytes wide, an IPv6 16 bytes wide.
+/// 8. addr_family: \[uint8\]. This array mirrors the addrs array. Each
+///    addr element will be mirrored in this array, except we'll write
+///    the associated IP addr family number. AF_INET for an IPv4,
+///    AF_INET6 for a v6.
```
leifwalsh, line 520: I think you need a 9, the canonical name, nul-terminated.

src/handlers.rs
```diff
@@ -134,6 +142,125 @@ pub fn handle_request(
...
+                Ok(hostent) => hostent,
+                Err(e) =>
+                // We shouldn't end up in that branch. Something
+                // got very very wrong on the glibc client side if
+                // we do. It's okay to bail, there's nothing much
+                // we can do.
+                {
+                    bail!("unexpected gethostbyaddr error: {}", e)
+                }
+            };
```
leifwalsh, line 208:
```suggestion
            let hostent = gethostbyaddr_r(LibcIp::V4(address_bytes))?;
```

I probably would have written this, and just let the original error surface and get logged. But I don't have strong (or generally good) opinions about error handling. I'll defer to @blinsay if he wants to have a stronger or better opinion.

(same idea below too)

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+                    for octet in ip6.octets() {
+                        buf_addrs.push(octet)
+                    }
+                }
+            }
+        }
+
+        for alias in self.aliases.iter() {
+            let alias_bytes = CString::new(alias.clone())?.into_bytes_with_nul();
+            let size_in_bytes: i32 = alias_bytes.len() as i32;
```
leifwalsh, line 420: Do you need the type twice?

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+        let has_addrs = num_addrs > 0;
+
+        let hostname_c_string_bytes = CString::new(self.name.clone())?.into_bytes_with_nul();
+        let hostname_c_string_len = if has_addrs {
+            hostname_c_string_bytes.len() as i32
+        } else {
+            0
+        };
+
+        let header = protocol::HstResponseHeader {
```
leifwalsh, line 440: I think this could be simplified by adding a `HST_RESPONSE_HEADER_NOT_FOUND` like you did for `AI_RESPONSE_HEADER_NOT_FOUND` - get rid of all the `has_addrs` checks. Unless `aliases` can still be valid or something?

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+            } else if num_v4 != 0 {
+                4
+            } else {
+                16
+            },
+            h_addr_list_cnt: num_addrs,
+            error: self.herrno,
+        };
+
+        let total_len = 4 * 8
```
leifwalsh, line 463: Where does `4 * 8` come from? Is that really just the size of `HstResponseHeader`?

src/handlers.rs
```diff
@@ -262,8 +384,194 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+        buf.extend_from_slice(header.as_slice());
+
+        // add hostname
+        if has_addrs {
+            buf.extend_from_slice(&hostname_c_string_bytes);
+
+            // Add aliases sizes
+            if !self.aliases.is_empty() {
+                buf.extend_from_slice(buf_aliases_size.as_slice());
+            }
```
leifwalsh, line 480:
```suggestion
            buf.extend_from_slice(buf_aliases_size.as_slice());
```

You can do this unconditionally, right? `buf_aliases_size` should be the empty vector if it's empty? (Same about `buf_aliases` below)

leifwalsh (changes requested):
A couple minor typos and some questions. Overall I think it's fairly straightforward (or as straightforward as code dealing with getaddrinfo and friends can be). Haven't tried running it at all but the code feels like it would be debuggable if something crashed?

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+#[derive(Clone, Debug)]
+pub struct Hostent {
+    pub name: String,
+    pub aliases: Vec<String>,
+    pub addr_type: i32,
+    pub addr_list: Vec<std::net::IpAddr>,
+    pub herrno: i32,
+}
+
+fn from_libc_hostent(value: libc::hostent) -> Result<Hostent, anyhow::Error> {
```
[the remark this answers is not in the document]
picnoir: Absolutely. Makes sense.

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+    let mut ret_hostent: libc::hostent = libc::hostent {
+        h_name: ptr::null_mut(),
+        h_aliases: ptr::null_mut(),
+        h_addrtype: 0,
+        h_length: 0,
+        h_addr_list: ptr::null_mut(),
+    };
+    let mut herrno: libc::c_int = 0;
+    let mut hostent_result = ptr::null_mut();
+    let mut buf: Vec<u8> = Vec::with_capacity(200);
```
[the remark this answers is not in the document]
picnoir: Empirical trial-and-error. I tried several gethostbyxxx requests and saw that most of them fit in a 200 bytes buffer. 

Looking [at](https://sourcegraph.com/github.com/NetBSD/src/-/blob/lib/libc/net/gethnamaddr.c?L1332) [sourcegraph](https://sourcegraph.com/github.com/NetBSD/src/-/blob/distrib/utils/libhack/gethost.c?L79) , it seems like there's no consensus wrt. the buffer size and we're in the lower range with 200 bytes (also, lots of ppl seem to disregard the ERANGE errno…).

Every time we resize the buffer, we have to go through another gethostbyxxx call, which is quite expensive. Better start with a size that fits most requests.

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+                &mut ret_hostent,
+                buf.as_mut_ptr() as *mut libc::c_char,
+                (buf.capacity() as size_t).try_into().unwrap(),
+                &mut hostent_result,
+                &mut herrno,
+            )
+        };
+
+        if ret == libc::ERANGE {
+            buf.reserve(buf.capacity() * 2);
```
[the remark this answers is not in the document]
picnoir: Good point. I guess that past 500M, we should be more than done. Let's use that arbitrary limit.

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+        } else {
+            break;
+        }
+    }
+    unmarshal_gethostbyxx(hostent_result, herrno)
+}
+
+///
```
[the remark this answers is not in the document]
picnoir: > What does "The stream" refer to?

I have honestly no idea what I mean here :O This + the extra line makes me think this comment has been butchered during a git conflict resolution at some point. Let's rewrite it.

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+/// The stream is positioned at the first entry in the directory.
+///
+/// af is nix::libc::AF_INET6 or nix::libc::AF_INET6
+pub fn gethostbyname2_r(name: String, af: libc::c_int) -> anyhow::Result<Hostent> {
+    let name = CString::new(name).unwrap();
+
+    // Prepare a libc::hostent and the pointer to the result list,
+    // which will passed to the ffi::gethostbyname2_r call.
+    let mut ret_hostent: libc::hostent = libc::hostent {
+        h_name: ptr::null_mut(), // <- points to buf
```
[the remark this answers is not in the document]
picnoir: It's nonsensical and outdated :)

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+        h_addrtype: 0,
+        h_length: 0,
+        h_addr_list: ptr::null_mut(),
+    };
+    let mut herrno: libc::c_int = 0;
+    let mut buf: Vec<u8> = Vec::with_capacity(200);
+    // We absolutely need to point hostent_result to the start of the
+    // result buffer. Glibc segfaults if we don't.
+    let mut hostent_result = ptr::null_mut();
+    let _: i32 = loop {
```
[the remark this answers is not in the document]
picnoir: We shouldn't, thanks.

(I rewrote this binding 3 times, we used to error out when the return value was !=0. Which is not what the nscd client expects…)

src/handlers.rs
```diff
@@ -134,6 +142,125 @@ pub fn handle_request(
...
         }
 
+        RequestType::GETAI => {
+            let hostname = CStr::from_bytes_with_nul(request.key)?.to_str()?;
+            // Boths hints are necessary to mimick the glibc behaviour.
+            let hints = AddrInfoHints {
+                // The canonical name will be filled in the first
+                // addrinfo struct returned by getaddrinfo.
+                flags: AI_CANONNAME,
+                // Deduplicate the resulting addrs.
```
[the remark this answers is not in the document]
picnoir: Clarified the comment:

```rust
                // Setting socktype to SOCK_STREAM to deduplicate the
                // results. If we don't do that, glibc gives us twice
                // every entry: once as a SOCK_STREAM, another one as
                // SOCK_DGRAM.
```

src/ffi.rs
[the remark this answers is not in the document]
flokli: I'd probably use 1..10M as an upper limit already. That's quite a lot of hostnames / IPs ;-)

src/ffi.rs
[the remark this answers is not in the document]
picnoir: On the Nscd side, the scratch buffer is initialized at 1024 bytes:

```
struct scratch_buffer {
  void *data;    /* Pointer to the beginning of the scratch area.  */
  size_t length; /* Allocated space at the data pointer, in bytes.  */
  union { max_align_t __align; char __c[1024]; } __space;
};
```

Let's re-use this default instead of this 200.

src/ffi.rs
```diff
@@ -40,3 +44,250 @@ pub fn disable_internal_nscd() {
...
+        h_aliases: ptr::null_mut(),
+        h_addrtype: 0,
+        h_length: 0,
+        h_addr_list: ptr::null_mut(),
+    };
+    let mut herrno: libc::c_int = 0;
+    let mut buf: Vec<u8> = Vec::with_capacity(200);
+    // We absolutely need to point hostent_result to the start of the
+    // result buffer. Glibc segfaults if we don't.
+    let mut hostent_result = ptr::null_mut();
```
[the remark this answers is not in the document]
picnoir: :+1:  Outdated comment.

src/handlers.rs
picnoir: Yeah, that's the header. Rewrote that more explicitly using `std::mem::size_of`

src/handlers.rs
picnoir: Added

```rust
                // There's no way to convey socktype in the the Nscd
                // protocol, neither in the request nor response.
                //
                // Set this to SOCK_STREAM to match glibc and unscd
                // behaviour.
```

Does it make sense?

picnoir:
I squashed the nits/small modifications in the original commits.

Separated the getaddrinfo serialization routine (asked in https://github.com/twosigma/nsncd/pull/71#discussion_r1375506047) in a separate commit: it's a lot, it'll likely be simpler to review it from a separate commit.

src/handlers.rs
picnoir: Did not address this one yet. Let me know what you prefer.

src/handlers.rs
picnoir: Added.

src/handlers.rs
flokli: @leifwalsh please mark as resolved if you're ok with this.

picnoir:
(fixed the clippy error)

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+        } else {
+            break;
+        }
+    }
+    unmarshal_gethostbyxx(hostent_result, herrno)
+}
+
+/// Typesafe wrapper around the gethostbyname2_r glibc function
+///
+/// af is either nix::libc::AF_INET6 or nix::libc::AF_INET6
```
leifwalsh, line 228:
```suggestion
/// af is either nix::libc::AF_INET or nix::libc::AF_INET6
```

leifwalsh (changes requested):

leifwalsh:
@picnoir have you sent us a [CLA](https://github.com/twosigma/nsncd/blob/main/cla/TSOS%20CLA%20-%20Individual%20-%20nsncd.pdf) yet? I'm trying to track this down internally too but people are on vacation/doing end-of-year activities so I figured I'd just ask you while I wait to hear back. If you haven't, please do so (emailing a scanned PDF is fine) and let me know.

leifwalsh:
Otherwise LGTM, @geofft @blinsay let me know if you want to take a pass, otherwise I'll merge once we have CLAs or on Wednesday, whichever comes first.

flokli:
> @flokli and @picnoir have you sent us a [CLA](https://github.com/twosigma/nsncd/blob/main/cla/TSOS%20CLA%20-%20Individual%20-%20nsncd.pdf) yet? I'm trying to track this down internally too but people are on vacation/doing end-of-year activities so I figured I'd just ask you while I wait to hear back. If you haven't, please do so (emailing a scanned PDF is fine) and let me know.
> 
> (I think we accepted one of @flokli's earlier PRs as de minimis, so we probably don't have one from either of you yet)

I'm very sure I sent one out Nov 14 2022 (to tsos at twosigma.com` just found the email), and am somewhat convinced @picnoir did so as well at around the same time.

picnoir:
Yup, according to https://github.com/twosigma/nsncd/discussions/44#discussioncomment-4135685 I did.

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+}
+
+/// This structure is the Rust counterpart of the `libc::hostent` C
+/// function the Libc hostent struct.
+///
+/// It's mostly used to perform the gethostbyaddr and gethostbyname
+/// operations.
+///
+/// This struct can be serialized to the wire through the
+/// `serialize` function or retrieved from the C boundary using the
```
flokli, line 86: serialize_hostent

src/handlers.rs
```diff
@@ -261,8 +387,203 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
     Ok(result)
 }
 
+impl Hostent {
+    fn serialize(&self) -> Result<Vec<u8>> {
```
flokli, line 391: To match the rest of the structure of this file, I'd probably change this to be a
```rust
fn serialize_hostent(hostent: Hostent) -> Result<Vec<u8>> {
  // …
}
```

src/handlers.rs
```diff
@@ -261,8 +387,203 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+///           implementation.
+/// 7. addrs: \[BE-encoded IPv4/IPv6\]. We sequentially write the
+///    IPv4 and IPv6 bytes using a big endian encoding. There's no
+///    padding, an IPv4 will be 4 bytes wide, an IPv6 16 bytes wide.
+/// 8. addr_family: \[uint8\]. This array mirrors the addrs array. Each
+///    addr element will be mirrored in this array, except we'll write
+///    the associated IP addr family number. AF_INET for an IPv4,
+///    AF_INET6 for a v6.
+/// 9. canon_name: Canonical name of the host. Null-terminated string.
+fn serialize_address_info(resp: &AiResponse) -> Result<Vec<u8>> {
```
flokli, line 533: This could also just take the `AiResponse` by value (consume it), so there's less cloning needed.

leifwalsh (approved):

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+) -> anyhow::Result<Hostent> {
+    if !hostent.is_null() {
+        let res = from_libc_hostent(unsafe { *hostent } )?;
+        Ok(res)
+    } else {
+        Ok(
+            // This is a default hostent header we're supposed to use
+            // to convey a lookup error. This is a glibc quirk, I have
+            // nothing to do with that, don't blame me :)
+            Hostent {
```
blinsay, line 174: want to move this to its own function? maybe something like `Hostent::error_value` would be nice

src/handlers.rs
blinsay: The whole thing with the sentinel error-case hostent getting returned from `gethostbyaddr_r` is very odd to me. We don't have to reproduce libc's strangeness exactly, even if we do have to serialize that error value and send it back to the caller.

What if we did something like:

```rust
enum HostentError {
  NotFound,
  TryAgain,
  NoRecovery,
  NoData,
  Other(anyhow::Error),
}

pub fn gethostbyname2_r(name: String, af: libc::c_int) -> Result<Hostent, HostentError> {
  ...
}
```

and then in handlers we could do something like this:

```rust
            let hostent = match gethostbyaddr_r(LibcIp::V4(address_bytes)) {
                Ok(hostent) => hostent,
                Err(Other(e)) => bail!("oh no, something very bad happened: {e}"),
                Err(e) => {
                  log::debug!("gethostbyaddr_r(..) = {e:?}");
                  Hostent::error_value()
                },
            };
```

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+        h_name: ptr::null_mut(),
+        h_aliases: ptr::null_mut(),
+        h_addrtype: 0,
+        h_length: 0,
+        h_addr_list: ptr::null_mut(),
+    };
+    let mut herrno: libc::c_int = 0;
+    // The 1024 initial size comes from the Glibc default. It fit most
+    // of the requests in practice.
+    let mut buf: Vec<u8> = Vec::with_capacity(1024);
```
blinsay, line 244: I think by the book this may need to be `MaybeUninit<u8>` even though it's never read. Let me consult with @geofft.

src/handlers.rs
```diff
@@ -261,8 +388,200 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+    }
+
+    for alias in hostent.aliases.iter() {
+        let alias_bytes = CString::new(alias.clone())?.into_bytes_with_nul();
+        let size_in_bytes = alias_bytes.len() as i32;
+        buf_aliases_size.extend_from_slice(&size_in_bytes.to_ne_bytes());
+        buf_aliases.extend_from_slice(alias_bytes.as_slice());
+    }
+
+    // this can only ever express one address family
```
blinsay, line 428: it's a bit confusing that this check happens after the aliases have been serialized.

src/handlers.rs
```diff
@@ -261,8 +388,200 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+        buf.extend_from_slice(buf_addrs.as_slice());
+
+        // add aliases
+        buf.extend_from_slice(buf_aliases.as_slice());
+
+        debug_assert_eq!(buf.len(), total_len);
+
+        buf
+    } else {
+        let error_header = protocol::HstResponseHeader {
```
blinsay, line 491: can this case get moved way up to the beginning of the function? it's a bit lost dangling down here:

```rust
if hostent.addr_list.is_empty() {
  return Ok(protocol::HstReponseHeader{ .... })
}

// rest of the fn goes here
```

src/handlers.rs
```diff
@@ -261,8 +388,200 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+///    AF_INET6 for a v6.
+/// 9. canon_name: Canonical name of the host. Null-terminated string.
+fn serialize_address_info(resp: AiResponse) -> Result<Vec<u8>> {
+    let mut b_families: Vec<u8> = Vec::with_capacity(2);
+    let mut b_addrs: Vec<u8> = Vec::with_capacity(2);
+    for addr in &resp.addrs {
+        match addr {
+            IpAddr::V4(ip) => {
+                b_families.push(AddressFamily::Inet as u8);
+                for octet in ip.octets() {
```
blinsay, line 539: nit: this can be `b_addrs.extend(ip.octets())`

src/handlers.rs
```diff
@@ -261,8 +388,200 @@ fn serialize_initgroups(groups: Vec<Gid>) -> Result<Vec<u8>> {
...
+                    b_addrs.push(octet);
+                }
+            }
+            IpAddr::V6(ip) => {
+                b_families.push(AddressFamily::Inet6 as u8);
+                for segment in ip.segments() {
+                    for byte in u16::to_be_bytes(segment) {
```
blinsay, line 546: any reason this doesn't use `octets`?

https://doc.rust-lang.org/stable/std/net/struct.Ipv6Addr.html#method.octets

blinsay (commented):
Thanks for taking three(!) passes at this, there's a ton here.

src/ffi.rs
blinsay: We chatted and I think I'm being overly cautious here.

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+
+    // ensure value.h_length matches what we know from this address family
+    if value.h_addrtype == libc::AF_INET && value.h_length != 4 {
+        bail!("unsupported h_length for AF_INET: {}", value.h_length);
+    }
+    if value.h_addrtype == libc::AF_INET6 && value.h_length != 16 {
+        bail!("unsupported h_length for AF_INET6: {}", value.h_length);
+    }
+
+    let name = unsafe { CStr::from_ptr(value.h_name).to_str().unwrap().to_string() };
```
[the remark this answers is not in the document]
flokli: Added a check.

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+        let mut aliases: Vec<String> = Vec::new();
+        let mut h_alias_ptr = value.h_aliases as *const *const libc::c_char;
+        while !(unsafe { *h_alias_ptr }).is_null() {
+            aliases.push(unsafe { CStr::from_ptr(*h_alias_ptr).to_str().unwrap().to_string() });
+            // increment
+            unsafe {
+                h_alias_ptr = h_alias_ptr.add(1);
+            }
+        }
+        // value.h_addrtype
```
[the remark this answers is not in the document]
flokli: Ack.

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+    if value.h_addrtype == libc::AF_INET && value.h_length != 4 {
+        bail!("unsupported h_length for AF_INET: {}", value.h_length);
+    }
+    if value.h_addrtype == libc::AF_INET6 && value.h_length != 16 {
+        bail!("unsupported h_length for AF_INET6: {}", value.h_length);
+    }
+
+    let name = unsafe { CStr::from_ptr(value.h_name).to_str().unwrap().to_string() };
+
+    Ok({
```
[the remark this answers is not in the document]
flokli: Ack.

src/ffi.rs
```diff
@@ -40,3 +44,248 @@ pub fn disable_internal_nscd() {
...
+        }
+        // value.h_addrtype
+
+        // construct the list of addresses.
+        let mut addr_list: Vec<std::net::IpAddr> = Vec::new();
+
+        // copy the pointer into a private variable that we can mutate
+        // h_addr_list is a pointer to a list of pointers to addresses.
+        // h_addr_list[0] => ptr to first address
+        // h_addr_list[1] => null pointer (end of list)
```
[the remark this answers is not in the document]
flokli: Updated the comments.

src/ffi.rs
flokli: Done, thanks for the suggestion.

src/handlers.rs
flokli: Moved before that.

src/handlers.rs
flokli: Moved that return further up, allowed simplifying some of the bottom, because we already ruled out these special cases.

src/handlers.rs
flokli: Yes, this shoudl work. Updated.

flokli:
> The whole thing with the sentinel error-case hostent getting returned from gethostbyaddr_r is very odd to me. We don't have to reproduce libc's strangeness exactly, even if we do have to serialize that error value and send it back to the caller.

We're just forwarding `herrno`, or setting it to `0` (in the case we documented in the last push). We don't get much from trying to interpret it and carving out specific enums.

Some iteration of PR tried to do this, but turned out to be very error-prone, for little added value. We opted to stick more to how it's done by glibc here, especially considering how much userspace relies on this to be identical.

blinsay:
> Some iteration of PR tried to do this, but turned out to be very error-prone, for little added value. We opted to stick more to how it's done by glibc here, especially considering how much userspace relies on this to be identical

What parts were more error prone? Is one of those iterations a closed PR I can go look at?

picnoir:
> Is one of those iterations a closed PR I can go look at?

Sure. https://github.com/nix-community/nsncd/blob/15133dd5bfbbc1111afb929a7005e8f672cdce68/src/ffi.rs#L56

Don't use that, it's not a sound approach.

> What parts were more error prone?

Trying to interpret herrno ourselves and return an error hostent if `herrno != 0`. The correct thing to do here is:

1. if `hostent != null`, set herrno to 0 and return the received hostent as it is.
2. if `hostent == null`, return the error header, forward herrno.

We're not logging anything about herrno, there's no reason trying to parse it. It can only yield to more errors.

In the grand scheme of things, Nscd is a glorified (and lossy) glibc proxy. We just forward what glibc throws at us. I think that architecturally-wise, we should try to parse as few things as possible. To be franc, I even came to believe we'd be better off not even trying to parse the string bytes into a Rust string. There are valid cases where a valid glibc string could be an invalid Rust string.

flokli:
@picnoir good call, I pushed another commit on top that now keeps these CStrings CStrings, allowing to remove some of the unnecessary back-and-forth parsing.

blinsay:
> I pushed another commit on top that now keeps these CStrings CStrings, allowing to remove some of the unnecessary back-and-forth parsing and serializing.

Nice :D

> We're not logging anything about herrno, we're not using it for any logic, there's no reason trying to parse it. It can only yield to more errors.
>
> In the grand scheme of things, Nscd is a glorified (and lossy) glibc proxy. We just forward what glibc throws at us. I think that architecturally-wise, we should try to parse as few things as possible. Just the bare minimum to perform the Nscd error logic.

This is the right position to take imo, but it was our choice to bundle herrno in with `libc::hostent` fields. It seems like`Hostent` contains an `herror` field because that was easy. Point taken on not parsing values we don't have to, but I think that means the right thing to do is just not interpret the error code, rather than pretend there's no difference between a `libc::hostent` and and error.

Given that I think I I'd suggest we do something like this:

```rust
enum HostentError {
  HError(i32),
  Other(anyhow::Error),
}
```

or even like this:

```rust
enum Hostent {
  Ok{ name: CString, ... },
  Error(i32),
}
```

picnoir:
done.

blinsay:
Awesome, looks good. I'm good to merge if @leifwalsh is.

leifwalsh (approved):

flokli:
:tada: thanks!

leifwalsh:
thank you both!

Issue #70, opened 2023-09-24
Fails to build if glibc is built without nscd.


micwoj92:
glibc build option is `--disable-nscd`

```
   Compiling libc v0.2.140
   Compiling cfg-if v1.0.0
   Compiling autocfg v1.1.0
   Compiling proc-macro2 v1.0.64
   Compiling unicode-ident v1.0.8
   Compiling quote v1.0.23
   Compiling crossbeam-utils v0.8.15
   Compiling slog v2.7.0
   Compiling time-core v0.1.0
   Compiling once_cell v1.17.1
   Compiling syn v1.0.109
   Compiling thread_local v1.1.7
   Compiling memoffset v0.7.1
   Compiling num-traits v0.2.15
   Compiling time-macros v0.2.8
   Compiling itoa v1.0.6
   Compiling anyhow v1.0.69
   Compiling num_threads v0.1.6
   Compiling slog-async v2.7.0
   Compiling crossbeam-channel v0.5.7
   Compiling static_assertions v1.1.0
   Compiling bitflags v1.3.2
   Compiling take_mut v0.2.2
   Compiling pin-utils v0.1.0
   Compiling sd-notify v0.4.1
   Compiling dirs-sys-next v0.1.2
   Compiling atty v0.2.14
   Compiling dirs-next v2.0.0
   Compiling atoi v2.0.0
   Compiling time v0.3.20
   Compiling term v0.7.0
   Compiling nix v0.26.2
   Compiling slog-term v2.9.0
   Compiling num-derive v0.3.3
   Compiling nsncd v1.4.1 (/build/nsncd-git/src/nsncd)
error: linking with `cc` failed: exit status: 1
  |
[...]

PR #94, opened 2024-02-13
Added support for libc built without nscd

Thesola10:
For reasons talked about over here https://github.com/twosigma/nsncd/issues/70#issuecomment-1942790292, this PR adds a non-default build feature, `glibc_no_nscd`, which allows `nsncd` to build on systems where the host `glibc` lacks `nscd` support, for use with e.g. programs installed using the Nix package manager.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -37,6 +37,7 @@ extern "C" {
 /// We _are_ nscd, so we need to do the lookups, and not recurse.
 /// Until 2.14, this function was taking no parameters.
 /// In 2.15, it takes a function pointer from hell.
+#[cfg(not(feature="glibc_no_nscd"))]
 unsafe extern "C" fn do_nothing(_dbidx: size_t, _finfo: *mut libc::c_void) {}
 
 /// Disable nscd inside our own glibc to prevent recursion.
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -47,7 +47,7 @@ pub fn disable_internal_nscd() {
         let __nss_disable_nscd =
             mem::transmute::<*mut libc::c_void, extern "C" fn(hell: unsafe extern "C" fn(size_t, *mut libc::c_void))>(sym_ptr);
 
-        if sym_ptr != ptr::null_mut() {
+        if !sym_ptr.is_null() {
             __nss_disable_nscd(do_nothing);
         }
     }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -15,20 +15,14 @@
  */
 
 use anyhow::anyhow;
-use nix::libc::{self};
+use nix::libc::{self, dlsym};
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
 use std::ptr;
+use std::mem;
 
 #[allow(non_camel_case_types)]
 type size_t = ::std::os::raw::c_ulonglong;
-
-#[cfg(not(feature="glibc_no_nscd"))]
-extern "C" {
-    /// This is the function signature of the glibc internal function to
-    /// disable using nscd for this process.
-    fn __nss_disable_nscd(hell: unsafe extern "C" fn(size_t, *mut libc::c_void));
-}
 
 /// Copied from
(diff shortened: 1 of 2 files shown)


leifwalsh (approved):
this seems fine, have you tested that it solves your problem?

Thesola10:
Thanks, yep this version solved my problem and Nix apps have access to usernames now!

Thesola10:
You may want to update your CI to not use `--all-features` by default

Thesola10:
Another solution would involve using `dlopen()` and `dlsym()`, to ensure the symbol is used at runtime in the event that the libc is replaced with a drop-in update:

```c
#define _GNU_SOURCE
#include <dlfcn.h>
#include <link.h>
#include <unistd.h>
#include <stdio.h>

int __nss_disable_nscd(void (*fh)(size_t dbidx, void *finfo))
{
    void *libc_handle;
    Dl_info dli;
    int (*_ndn)(size_t dbidx, void *finfo);

    dladdr1((void *) read, &dli, libc_handle, RTLD_DL_LINKMAP);
    _ndn = dlsym(libc_handle, "__nss_disable_nscd");

    if (_ndn)
        _ndn(dbidx, finfo);
}
```

This solution relies on the fact that `struct link_map` _is_ the handle format used by `dlsym()`, so we "get info" on a known `libc` symbol, to obtain the handle on our own `libc` and find the real function.

This is probably reasonably simple to transcribe to unsafe Rust, but may not work on uClibc from what I can gather, so this approach itself would have to be locked behind a feature -- although one that can be enabled regardless of host `nscd` support.

Thesola10:
Another solution would involve using `dlsym()`, to ensure the symbol is used at runtime in the event that the libc is replaced with a drop-in update. It would look something like this draft:

```c
#define _GNU_SOURCE
#include <dlfcn.h>
#include <link.h>
#include <unistd.h>
#include <stdio.h>

int __my_nss_disable_nscd(void (*fh)(size_t dbidx, void *finfo))
{
    int (*_ndn)(size_t dbidx, void *finfo);

    _ndn = dlsym(NULL, "__nss_disable_nscd");

    if (_ndn)
        _ndn(dbidx, finfo);
}
```

This is probably reasonably simple to transcribe to unsafe Rust.

src/ffi.rs
```diff
@@ -15,42 +15,46 @@
...
-        __nss_disable_nscd(do_nothing);
+        let sym_name = CString::new("__nss_disable_nscd").unwrap();
+        let sym_ptr = dlsym(ptr::null_mut(), sym_name.as_ptr());
+        let __nss_disable_nscd =
+            mem::transmute::<*mut libc::c_void, extern "C" fn(hell: unsafe extern "C" fn(size_t, *mut libc::c_void))>(sym_ptr);
+
+        if sym_ptr != ptr::null_mut() {
+            __nss_disable_nscd(do_nothing);
+        } else {
+            println!("No nscd support, hum dee dum...");
```
Thesola10, line 53:
```suggestion
```

Huh, forgot to drop the debug message.

Thesola10:
This new approach has been tested on both Arch (no `nscd` support) and NixOS (with `nscd` support) with all unit tests passed with flying colors

leifwalsh:
`dlsym` sounds like a better, cfg-free approach, but now clippy's mad. 

Could you also send us a CLA? Emailed is fine: https://github.com/twosigma/nsncd/blob/main/.github/CONTRIBUTING.md

Thesola10:
Done! Check that my email `me <at> thesola.io` doesn't somehow end up in spam.

leifwalsh:
@Thesola10 I think that should be the same, mind testing one more time to be sure?

leifwalsh (approved):

Thesola10:
Looks good, works fine, thanks

leifwalsh:
@Thesola10 we don't really have a release schedule (since we're not on crates.io) but if you need us to cut a tag for some reason, let us know. Otherwise we generally only do that when we need to do an internal release.

PR #110, opened 2024-04-23
nsncd.service: run as non-root

### LINKED ISSUE #14: Run as non-root
It occurs to me there's no particular reason to keep running as root (at least on most systems, though for certain nss_ldap configurations, you've got a password that's readable only as root). We should support dropping privileges.

I think best practice is to run as your own system account instead of `nobody` because that quickly turns `nobody` into a fairly juicy target of its own, so probably this wants to be an argument to drop privileges to a specific user + some packaging config to create a system user.

flokli:
This updates the nsncd.service systemd unit file to run nsncd as a `nscd` user and group, and enables sandboxing.

Fixes #14.


flokli:
@leifwalsh I'm not sure if something on the debian side needs updating, like, ensuring the `nscd` user exists, via some hooks, as per https://www.debian.org/doc/manuals/securing-debian-manual/bpp-lower-privs.en.html. It seems like there's no `nscd` user in `base-passwd`.

Can you take a look at that?

flokli:
I got an email about merging in `main` here. Still keen on seeing this merged.

Looks like some packages for Debian 11 disappeared. Do you still need to support Debian 11, or can these parts be removed from the pipeline?

morisja:
Hi @flokli - yes I noticed deb11 issues this weekend, I got some time to execute some maintenance and have a pr to remove deb11 support, along with some other misc fixes. I want to implement the change you have PR'd as well. Once I get the maintenance changes merged I will look at this, thanks and apologies for the delay

morisja:
Made a release with all the recent changes, tested your unit changes ( and some other attributes) at work successfully - onto review hope to merge next week 👍

PR #111, opened 2024-04-29
add support for socket activation

flokli:
If nsncd is started using socket activation, use the passed FD instead of manually binding on SOCKET_PATH.

This doesn't change any current behaviour in the currently documented startup mode, but makes nsncd fit to run in a socket-activated environment.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -53,6 +53,7 @@ use std::time::Duration;
 
 use anyhow::{Context, Result};
 use crossbeam_channel as channel;
+use listenfd::ListenFd;
 use sd_notify::NotifyState;
 use slog::{debug, error, o, Drain};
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -77,19 +78,32 @@ fn main() -> Result<()> {
     let logger = slog::Logger::root(drain, slog::o!());
 
     let config = Config::from_env()?;
-    let path = Path::new(SOCKET_PATH);
+
+    // If we're started using socket activation, use the unix listener at index 0,
+    // else bind manually on SOCKET_PATH.
+    let (listener, listen_address) = match ListenFd::from_env()
+        .take_unix_listener(0)
+        .expect("invalid socket type at index")
```
blinsay, line 86: I think it makes slightly more sense to log an error and exit here but I don't feel strongly about it.

src/main.rs
```diff
@@ -77,19 +78,32 @@ fn main() -> Result<()> {
...
+    {
+        Some(listener) => (listener, "sd-listen-unix"),
+        None => {
+            let path = Path::new(SOCKET_PATH);
```
blinsay, line 90: let's throw this in a `fn`

blinsay (changes requested):
Looks good, some comments in line. Have you tested it locally with socket activation yet?

flokli:
> Looks good, some comments in line. Have you tested it locally with socket activation yet?

I did check it actually starting up, via `systemd-socket-activate`:

```
❯ systemd-socket-activate -l /tmp/foobar.sock target/debug/nsncd
Listening on /tmp/foobar.sock as 3.
Communication attempt on fd 3.
Execing target/debug/nsncd (target/debug/nsncd)
```

I wrote some spam to it, pointing curl to it, and (successfully) got back an error about the wrong protocol version:

```
May 08 11:43:02.917 DEBG accepted connection, stream: UnixStream { fd: FileDesc(OwnedFd { fd: 4 }), local: "/tmp/foobar.sock" (pathname), peer: (unnamed) }, thread: worker_2
May 08 11:43:02.917 DEBG parsing request, err: wrong protocol version 542393671, thread: worker_2
```

I have some systemd config to wire it up there, but I ran into some unrelated other issues and didn't get to fully test it yet. I will, happy to mark it as draft until then.

flokli:
Shelving these efforts for now, let's revisit this once https://github.com/systemd/systemd/pull/30360 has landed in a release, as it'd probably be revisited after that anyways.

PR #126, opened 2024-10-02
120 - Support service and netgroup lookups

morisja:


leifwalsh (approved):
I think FromStr is less in vogue than I remembered but I'll try switching it to From<&str> in a bit

PR #133, opened 2024-10-10
Issue 123 - enhance CI tests

morisja:
Simulate the use case where the nscd socket is mounted into a container.

Debootstrap a minimal debian install, start nsncd, run tests via systemd-nspawn with the socket mounted.


morisja:
Will merge with the reentrant function PR

PR #135, opened 2024-10-10
Issue 134 ,123- Enhance CI, Use reentrant functions for service lookups

morisja:


src/handlers.rs
```diff
@@ -107,25 +125,30 @@ impl FromStr for ServiceWithName {
 
 impl ServiceWithName {
     fn lookup(&self) -> Result<Option<Service>> {
-        let serv_entry: *const servent;
+        let service_cstr = CString::new(self.service.as_str())?;
+        let proto_cstr = self.proto.as_ref().map(|s| CString::new(s.as_str()).unwrap());
```
leifwalsh, line 129: `unwrap()` can panic, we should return an error instead

src/handlers.rs
```diff
@@ -107,25 +125,30 @@ impl FromStr for ServiceWithName {
...
-                    self.service.as_ptr() as *const c_char,
-                    protocol.as_ptr() as *const c_char,
-                )
-            };
-        } else {
-            serv_entry =
-                unsafe { getservbyname(self.service.as_ptr() as *const c_char, std::ptr::null()) };
-        }
+        let mut result_buf: servent = unsafe { mem::zeroed() };
+        let mut buffer: Vec<c_char> = vec![0; 1024];
```
leifwalsh, line 134: Is this always large enough?

leifwalsh (approved):

PR #138, opened 2024-10-11
factor erange handling into its own function

leifwalsh:


morisja (approved):

picnoir:
Nice. FTR, we have a few occurences of this pattern in src/ffi.rs.

morisja:
ah I should prob move a number of the functions I just added for service and netgroup to ffi.rs.

PR #144, opened 2024-10-22
Correctly handle lookups of missing netgroups

morisja:
If the netgroup does not exist, setnetgrent returns 0. Do not bail, simply return 0 results

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -342,9 +342,9 @@ impl NetgroupWithName {
         };
 
         if unsafe { setnetgrent(netgroup_name.as_ptr() as *const c_char) } != 1 {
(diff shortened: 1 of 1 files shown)


leifwalsh (approved):

src/handlers.rs
```diff
@@ -342,9 +342,9 @@ impl NetgroupWithName {
         };
 
         if unsafe { setnetgrent(netgroup_name.as_ptr() as *const c_char) } != 1 {
```
leifwalsh, line 344: wait, does it ever return anything but 0 or 1? do we need to check errno?

morisja: I was curious about that too, this mentioned only 1 or 0:
https://man7.org/linux/man-pages/man3/setnetgrent.3.html

PR #171, opened 2025-07-28
Release 1.5.2

morisja:


leifwalsh (approved):

picnoir:
nit: not a massive deal, but this release tag `v.1.5.2` broke the `v$version` pattern of the previous releases with thre addition of the dot after the v.

flokli:
I guess there should be no problem to also tag the same commit with `v1.5.2` as well (and maybe delete the old tag considering it's still quite unlikely it has been widely packaged).

morisja:
Ugh sorry about that, updated.

PR #182, opened 2025-10-12
Remove vendor rules target

morisja:
Builds started to fail recently on all debian versions.

The closest reported issue I could find was on openwrt; related to a rustlang change, both are referenced below.

As we no longer build on debian 9 (minimum 11) - I removed the vendor rules target and special handling

https://github.com/openwrt/openwrt/issues/20189

https://github.com/rust-lang/cargo/commit/29ecb7f36a963bed4729bd6909200b4302a366a5

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,6 @@ jobs:
           toolchain: ${{ github.event.schedule && 'nightly' || 'stable' }}
       - name: Build package
         run: |
-          debian/rules vendor
           dpkg-buildpackage --no-sign
         working-directory: clone
       - uses: actions/upload-artifact@v4
@@ -128,7 +127,6 @@ jobs:
           toolchain: ${{ github.event.schedule && 'nightly' || 'stable' }}
       - name: Build package
         run: |
-          debian/rules vendor
           dpkg-buildpackage --no-sign
         working-directory: clone
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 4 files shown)


morisja:
I think this is the correct thing to do  (was missing history on why we needed to patch a single crate locally) - but appears to build cleanly on all dists.

geofft:
Does this PR build successfully internally? It looks like the CI jobs are accessing crates.io over the network, which is what the `debian/vendor` thing was trying to avoid, and I think that will fail internally.

Or are you using the binary .debs from the public GHA builds and so that isn't a problem for you? If so I think it's fine to drop the vendoring hack entirely, but I'd probably leave a README.source in there saying that the build expects to run with internet access.

It's possible that newer versions of Debian have enough packaged Rust libraries to be able to add then to debian/control as build deps, which would allow the build to happen offline. You might want to take a look at whatever the official Debian packaging for nsncd ended up doing for debian/.

morisja:
Ahh thanks Geoff, that gives a lot of background. I've been using the deb built here but will test internally.
Will test this, try including the debian packages in control, take a look at salsa and update the README.source appropriately.

geofft:
It's totally fine if you want to decide to stop using the internal build and just use the .debs from here! The git history is kind of confusing. I think I put this setup together before e.g. we were using the GHA builds of twosigma/openjdk.

leifwalsh (dismissed):

morisja:
I took a look at the debian build, the vendor rule and .github directory contents that invoke it are present, but I don't think they are used. nsncd is only present in trixie, as such I don't think they run into the issues we are seeing here as they only build on a single distribution. Their debian/control includes regular pre-packaged build deps. I tested their build deps and can get deb 12 and 13 builds working, 11 does not have the required deps available.

We will use the debs built here, and I will switch debian/control build-deps to use the in distribution deps when 11 is fully retired.

leifwalsh (approved):

PR #203, opened 2026-05-30
Harden the shipped systemd service

### LINKED ISSUE #156: Harden nsncd.service
Where the authors are probably most familiar with what access is required, it would be nice if the `nsncd.service` file was updated with a good and secure default hardening setup.

When the service is running on NixOS there is _some_ hardening applied, but the service still gets a rating of `8.2 EXPOSED 🙁` in `systemd-analyze security nscd.service`. The existing hardening [is available here](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/system/nscd.nix#L138) ([here's another users config](https://github.com/rafaelsgirao/nixos-config/blob/main/modules/core/hardening.nix#L136)), and is obviously not very strict.

There is [an issue in the Nix repo](https://github.com/NixOS/nixpkgs/issues/377827) focused on hardening system services. While hopefully someone will pick up `nsncd` it would probably be better if the info was available here.

jeet-dekivadia:
## Summary
- empty the service capability bounding set and isolate kernel-facing interfaces that NSS modules should not need
- restrict namespace, realtime, personality, and alternate-ABI use while preserving host configuration, network access, and helper-daemon communication
- deny clock, raw-I/O, reboot, and swap syscall groups without changing service identity, filesystem visibility, or socket creation

Fixes #156

## Validation
- `git diff --check`
- reviewed the selected directives against the NSS compatibility constraints discussed in #156

## Remote Linux validation
- `git diff --check`
- `systemd-analyze verify nsncd.service`

### DIFF
--- a/nsncd.service
+++ b/nsncd.service
@@ -17,7 +17,22 @@ Description=name-service non-caching daemon
 
 [Service]
 ExecStart=/usr/lib/nsncd
+# NSS modules may need host configuration, network access, and helper daemons.
+# Restrict kernel-facing interfaces without isolating those dependencies.
+CapabilityBoundingSet=
+LockPersonality=true
+PrivateDevices=true
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHostname=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
 Restart=always
+RestrictNamespaces=true
+RestrictRealtime=true
+SystemCallArchitectures=native
+SystemCallFilter=~@clock @raw-io @reboot @swap
 Type=notify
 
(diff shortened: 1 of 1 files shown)


jeet-dekivadia:
Small follow-up for review context. This is deliberately service-boundary hardening rather than a behavior change: it blocks alternate ABIs, namespace creation, realtime scheduling, high-risk syscall groups, and kernel-facing interfaces while preserving host configuration, network access, helper-daemon communication, service identity, and socket creation.

Validation from preparation: git diff --check; systemd-analyze verify nsncd.service; reviewed the selected directives against the NSS compatibility constraints in #156 and systemd.exec semantics.

PR #204, opened 2026-05-30
Reject invalid request key lengths before slicing

jeet-dekivadia:
## Summary

- convert the signed protocol key length to `usize` before offset arithmetic
- reject negative and overflowing request key lengths as parser errors
- add regressions for negative and oversized wire lengths

The socket protocol carries `key_len` as an `i32`. Previously, values from `-1` through `-12` produced an end offset before the key start and then panicked while slicing. Large values could also overflow the signed addition in checked builds. Since the daemon socket is intentionally available to local clients, malformed requests should be rejected without taking down a worker.

## Validation

- `git diff --check`
- `rustfmt --edition 2018 --check src/protocol.rs`

## Remote Linux validation
- `git diff --check`
- `rustfmt --edition 2018 --check src/protocol.rs`
- `cargo test` (31 passed, 1 ignored)
- `cargo clippy --bins --tests -- -D warnings`

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -99,7 +99,10 @@ impl<'a> Request<'a> {
             .with_context(|| format!("invalid enum value {type_val}"))?;
 
         let key_len = buf[8..12].try_into().map(i32::from_ne_bytes)?;
-        let key_end = (12 + key_len).try_into()?;
+        let key_len: usize = key_len.try_into().context("invalid key length")?;
+        let key_end = 12usize
+            .checked_add(key_len)
+            .context("request key length overflow")?;
         ensure!(buf.len() >= key_end, "request body too small");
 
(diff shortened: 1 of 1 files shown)


jeet-dekivadia:
Small follow-up for review context. This keeps the malformed-request handling on the safe side of the parser boundary: signed key lengths are rejected before offset arithmetic and slicing, so a negative or overflowing wire length cannot panic a worker while handling a local socket request.

Validation from preparation: git diff --check; rustfmt --edition 2018 --check src/protocol.rs; cargo test, with 31 passed and 1 ignored; cargo clippy --bins --tests -- -D warnings. The patch adds regressions for both negative and oversized wire lengths.

leifwalsh (approved):

PR #212, opened 2026-09-03
Issue 206

### LINKED ISSUE #206: not-found group/passwd responses have `version=0`, breaking musl clients (EIO)
nsncd's not-found response for `GETGRBYNAME`/`GETGRBYGID` (and `GETPWBY*`) sets the header's `version` field to **0** instead of `VERSION` (2), because the not-found branch serializes a `#[derive(Default)]` header:

- [`handlers.rs#L363`](https://github.com/nix-community/nsncd/blob/e20cd397833f880c9135ea5b327b05d59aa17467/src/handlers.rs#L363) — group not-found: `let header = protocol::GrResponseHeader::default();`
- [`protocol.rs#L150`](https://github.com/nix-community/nsncd/blob/e20cd397833f880c9135ea5b327b05d59aa17467/src/protocol.rs#L150) — `#[derive(..., Default)] struct GrResponseHeader { version: c_int, ... }` → `version` defaults to `0`
- [`protocol.rs#L36`](https://github.com/nix-community/nsncd/blob/e20cd397833f880c9135ea5b327b05d59aa17467/src/protocol.rs#L36) — `pub const VERSION: i32 = 2;`

glibc's nscd client tolerates this, but **musl** validates the version field *before* `found` and rejects it with **EIO**:

- [musl `nscd_query.c#n106`](https://git.musl-libc.org/cgit/musl/tree/src/passwd/nscd_query.c#n106) — `if (buf[0] != NSCDVERSION) { errno = EIO; goto error; }` (`NSCDVERSION == 2`)

So on a host running nsncd, a static-musl binary looking up **any name absent from `/etc/group`/`/etc/passwd`** gets `EIO` instead of a clean not-found. (musl reads files first and only queries nscd on a miss, so existing names are unaffected — only the not-found path hits this.)

## Repro

`libc = "0.2"`, with nsncd serving `/run/nscd/socket`:

```rust
fn main() { unsafe {
    for n in ["users", "no_such_group_xyz"] {
        let c = std::ffi::CString::new(n).unwrap();
        *libc::__errno_location() = 0;
        let g = libc::getgrnam(c.as_ptr());
        println!("{n}: null={} errno={}", g.is_null(), *libc::__errno_location());
    }
}}
```

```
cargo run                                     # glibc: no_such_group_xyz -> null=true errno=0
cargo run --target x86_64-unknown-linux-musl  # musl:  no_such_group_xyz -> null=true errno=5 (EIO)
```

**Env:** nsncd 1.5.2, NixOS 26.11, static `x86_64-unknown-linux-musl` (rust-std 1.97.1), glibc 2.42.

## Fix

Set the version in the not-found/error headers — [`handlers.rs#L363`](https://github.com/nix-community/nsncd/blob/e20cd397833f880c9135ea5b327b05d59aa17467/src/handlers.rs#L363) and the `PwResponseHeader::default()` at [`handlers.rs#L319`](https://github.com/nix-community/nsncd/blob/e20cd397833f880c9135ea5b327b05d59aa17467/src/handlers.rs#L319):

```rust
let header = protocol::GrResponseHeader { version: protocol::VERSION, ..Default::default() };
```

or give the response headers a hand-written `Default` that sets `version: VERSION`.

morisja:
Fix version in not-found passwd and group responses

The default NSCD response headers have version 0 because Rust's
derived Default implementation initializes the version field to zero.

glibc tolerates this, but musl validates the NSCD response version
before checking the found field and returns EIO for these responses.

Set the protocol version explicitly in the not-found passwd and group
responses so that they contain a valid NSCD v2 header.

Fixes #206

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -200,9 +200,7 @@ fn env_positive_usize(var: &str, default: usize) -> Result<usize> {
         Ok(s) => s,
         Err(_) => return Ok(default),
     };
-    let val = s
-        .parse()
-        .with_context(|| format!("parsing int from {s}"))?;
+    let val = s.parse().with_context(|| format!("parsing int from {s}"))?;
     if val > 0 {
         Ok(val)
     } else {
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -855,7 +855,11 @@ fn serialize_service(service: Option<Service>) -> Result<Vec<u8>> {
             result.extend_from_slice(alias_bytes);
         }
     } else {
-        let header = protocol::ServResponseHeader::default();
+        let header = protocol::ServResponseHeader {
+            version: protocol::VERSION,
+            ..Default::default()
+        };
+
         result.extend_from_slice(header.as_slice());
     }
     Ok(result)
@@ -893,7 +897,10 @@ fn serialize_user(user: Option<User>) -> Result<Vec<u8>> {
         result.extend_from_slice(dir_bytes);
         result.extend_from_slice(shell_bytes);
     } else {
-        let header = protocol::PwResponseHeader::default();
+        let header = protocol::PwResponseHeader {
+            version: protocol::VERSION,
+            ..Default::default()
+        };
         result.extend_from_slice(header.as_slice());
     }
     Ok(result)
@@ -936,7 +943,10 @@ fn serialize_group(group: Option<Group>) -> Result<Vec<u8>> {
             result.extend_from_slice(member_bytes);
         }
     } else {
-        let header = protocol::GrResponseHeader::default();
+        let header = protocol::GrResponseHeader {
+            version: protocol::VERSION,
+            ..Default::default()
+        };
         result.extend_from_slice(header.as_slice());
     }
     Ok(result)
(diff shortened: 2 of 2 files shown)


geofft:
Presumably we also want to do this with `let header = protocol::ServResponseHeader::default();` earlier in the same file?

Also, looking very quickly, for `serialize_address_info` (`GETAI`) we serialize a constant `AI_RESPONSE_HEADER_NOT_FOUND` which has the right version. I wonder if we want to make the handling consistent, either doing the same thing with `..default()` or making constants for all the other not-found responses.

I'm also thinking a little bit about how to make this more foolproof for other response types in the future. One possible idea that I haven't fully thought through: make the `version` an enum with a single member with the right value, so that `::default()` works correctly, e.g.,

```rust
#[derive(Clone, Copy, Default)]
#[repr(u32)]
pub enum VERSION {
    #[default]
    VERSION = 2,
}

#[repr(C)]
#[derive(Clone, Copy, Default)]
pub struct PwResponseHeader {
    pub version: VERSION,
    // ...
}
```

.github/workflows/test.yml
```diff
@@ -166,7 +167,7 @@ jobs:
 
   run-ci-ubuntu-latest:
     runs-on: ubuntu-latest
-    needs: [build-debian-package-11, build-debian-package-12, build-debian-package-13]
+    needs: [build-debian-package-11, build-debian-package-12, build-debian-package-13, build-musl]
```
leifwalsh: Need to add this job too, right?

morisja: Thanks, was a little optimistic on the build side, I'll add a musl specific test instead

morisja:
Ah thanks Geoff, I'll take a look at these points!

morisja:
@geofft - I've made the update for ServResponseHeader, spent some time understanding your other feedback and have a couple of options (the one your proposed and adding a named not_found constructor to the header impl - I'll put some options together in separate branches and tag you, thanks for the tips!

Sample for default impl per header
https://github.com/twosigma/nsncd/pull/215

leifwalsh (approved):

PR #213, opened 2026-09-03
Add arm64 build and test, retire debian 11

morisja:
I'm interested in arm64 and debian 11 is now retired.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,6 +65,24 @@ jobs:
       - name: Build
         run: cargo build --verbose
 
+
+  build-debian-13-arm64:
+    runs-on: ubuntu-24.04-arm
+    container: debian:trixie
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install dependencies
+        run: |
+          apt-get update
+          apt-get -y install build-essential dpkg-dev ca-certificates sudo curl
+          apt-get -y build-dep .
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ github.event.schedule && 'nightly' || 'stable' }}
+      - name: Build
+        run: cargo test --verbose
+
+
   clippy:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -63,7 +63,8 @@ jobs:
         with:
           toolchain: ${{ github.event.schedule && 'nightly' || 'stable' }}
       - name: Build
-        run: cargo build --verbose
+        run: cargo test --verbose
+
```
leifwalsh, line 67: Why this change?

.github/workflows/test.yml
```diff
@@ -159,30 +161,30 @@ jobs:
...
         run: ci/test.sh
         env:
           HAVE_SYSTEMD: "1"
 
-
-  run-ci-debian-11:
-    runs-on: ubuntu-latest
-    container: debian:11
-    needs: [build-debian-package-11]
+  run-ci-unbuntu-24-04-container-arm64:
```
leifwalsh:
```suggestion
  run-ci-ubuntu-24-04-container-arm64:
```

.github/workflows/test.yml
```diff
@@ -192,12 +194,29 @@ jobs:
...
+          name: deb-package-debian-13-arm64
       - name: CI
         run: ci/test.sh
         env:
           HAVE_SYSTEMD: "0"
 
+
+  run-ci-ubuntu-latest:
+    runs-on: ubuntu-latest
+    needs: [build-debian-package-12, build-debian-package-13]
```
leifwalsh: Why does this depend on both 12 and 13?

.github/workflows/test.yml
```diff
@@ -236,10 +255,28 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install build-essential systemd-container debootstrap
+      - uses: actions/download-artifact@v8
+        with:
+          name: deb-package-debian-12
+      - name: CI
+        run: ci/test_nspawn.sh
+
+  run-enhanced-ci-ubuntu-24-04-arm64:
```
leifwalsh, line 277: Some of these jobs aren't depended on, should we add them all to the dependencies of `create-release`?

leifwalsh (changes requested):

.github/workflows/test.yml
morisja: Unintentional

.github/workflows/test.yml
morisja: resolved

.github/workflows/test.yml
morisja: yes!

leifwalsh (approved):

PR #215, opened 2026-09-05
Add default impls for all header responses

morisja:
Follow up to https://github.com/twosigma/nsncd/pull/212 based on discussion of https://github.com/twosigma/nsncd/issues/206

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -735,9 +735,9 @@ fn serialize_innetgr(innetgr: bool) -> Result<Vec<u8>> {
 
     if innetgr {
         let header = protocol::InNetgroupResponseHeader {
-            version: protocol::VERSION,
             found: 1,
             result: 1,
+            ..Default::default()
         };
         result.extend_from_slice(header.as_slice());
     }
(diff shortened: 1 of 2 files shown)


src/handlers.rs
```diff
@@ -995,14 +1000,15 @@ fn serialize_hostent(hostent: Hostent) -> Result<Vec<u8>> {
...
             protocol::HstResponseHeader {
-                version: protocol::VERSION,
                 found: 0,
                 h_name_len: 0,
                 h_aliases_cnt: 0,
                 h_addrtype: -1,
                 h_length: -1,
                 h_addr_list_cnt: 0,
                 error: hostent.herrno,
+                            ..Default::default()
```
leifwalsh: Formatting here looks weird to me?

leifwalsh (changes requested):

src/handlers.rs
morisja: huh so it was, I thought I ran cargo fmt earlier, should add that somewhere - resolved and updated

leifwalsh (approved):

